Issue #4, opened 2017-01-06
Support for openhab 1.x


paphko:
The current implementation only supports items with tags which are not available in OH1. I would suggest the following additional configuration for making items visible to the skill: dedicated groups. Groups are available in OH1 and OH2, so this solution should be independent of the server version.

A preconfigured group could be 'gAlexa' or simply 'Alexa'. When this group is configured in config.js, it can also be changed as desired. We can even allow a configuration of multiple groups so that all items in all groups are collected.
Furthermore, the configured groups could also reflect the supported [device types of smarthome skills](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/smart-home-skill-api-reference). Examples:

```
module.exports = {
    //userpass: 'user@foo.com:Password1',
    host: 'my.openhab.org',
    port: 443,
    path: '/rest/items/',

    generalGroups: ['Alexa'], // item type determines capabilities, e.g. percent for dimmers
    switchGroups: ['AlexaLights', 'AlexaSockets'], // allows on/off messages
    temperatureGroups: ['AlexaThermostats'], // allows temperature messages
    percentageGroups: ['AlexaDimmers', 'AlexaRollershutters'] // allows percentage messages
};
```

What do you think? I already tested switches with OH1 and it seems to work pretty well.

As part of this implementation, the code should be refactored to eliminate duplicate code (oh1.js and oh2.js are almost identical).

paphko:
I mixed up REST API of OH1 and OH2 - it seems as if the REST API of OH1 does not provide information which items are contained in a group :-(

paphko:
In case someone wants to try it out, too: https://github.com/paphko/openhab-alexa/tree/oh1_oh2_groups

I'll test this a few days and if it's stable, I'll create a PR.

paphko:
Although this solution works, it has the drawback that the (technical) item names must be used as device names for voice recognition because the OH1 REST API does not provide a label for items. So if you want to use this Skill, you should set proper item names.

I think a PR is not needed because OH2 should replace OH1 quite soon.

htreu:
A solution for OH1 might be to load the items from a special sitemap. This way you can provide separate item labels for voice control. The [OpenHAB-Homekit-Bridge](https://github.com/htreu/OpenHAB-HomeKit-Bridge) is implemented this way.

paphko:
You're right, a sitemap would have been a more elegant solution :-)
[...]

PR #8, opened 2017-01-20
Remove OH1 because it doesn't work with current implementation (Homekit tags)

paphko:
The current implementation uses Homekit tags which does not work with OH1 because it does not support tags. And since OH2 is intended to replace OH1, I suggest to remove OH1-related code entirely.

I successfully tested [this fork](https://github.com/paphko/openhab-alexa/tree/oh1_oh2_groups) which uses a dedicated group to bind items to Alexa, but this has further drawbacks (the item name must be used  as device name, which is typically rather technical and not suited for voice recognition; the reason is that labels are also not supported in OH1 REST API). So if there is someone who wants to use this Skill for OH1, the aforementioned fork can be used.


digitaldan:
Hi Paphko, thanks for the feedback and work.  I think we should remove it as well.  I left it in there just in case someone else wanted to play with it, but as you have discovered it's difficult to support in a standard way on OH1.  Also since this is exactly the code we will be running for the official openHAB published skill, I think this project should not contain functionality we would not support there.

digitaldan:
I am going to go ahead and merge this, thanks!

cole701:
Does removing the OH1 support with this issue also then close out issue #4 which requests OH1 support or did I read it wrong?

digitaldan:
I would say so, I will go ahead and close.

PR #19, opened 2017-03-26
Lock support

fourthwall:
Hi,

I attempted to add support for the Locking system in alexa - mapping to a switchable. 

This is for US users only so I can't test it, but if any US users that are more experienced want to expand on this, that'd be nice. It looks OK to me - although the nested functions in the setLockState function were guesswork a bit. Should work.

It tries to check if the switch actually updated or not when confirming the lock response - I probably got this a bit wrong. But nonetheless it's here for you to view - and would form a basis for how a lock would likely work in OpenHAB - as a switchable that does not update until it has confirmed the lock has responded.

### DIFF
--- a/oh2.js
+++ b/oh2.js
@@ -548,10 +548,10 @@ function discoverDevices(token, success, failure) {
                   var additionalApplianceDetails = {};
                   switch (tag) {
                   case "Locks":
-                    actions = [
-                        "GetLockStateRequest",
-                        "SetLockStateRequest"
-                    ];
+                      actions = [
+                          "GetLockStateRequest",
+                          "SetLockStateRequest"
+                      ];
                     break;
                   case "Lighting":
                   case "Switchable":
--- a/oh2.js
+++ b/oh2.js
@@ -547,6 +547,12 @@ function discoverDevices(token, success, failure) {
                   var actions = null;
                   var additionalApplianceDetails = {};
                   switch (tag) {
+                  case "Locks":
+                    actions = [
+                        "GetLockStateRequest",
+                        "SetLockStateRequest"
+                    ];
+                    break;
                   case "Lighting":
                   case "Switchable":
                       actions = getSwitchableActions(item);
--- a/oh2.js
+++ b/oh2.js
@@ -498,7 +498,7 @@ function setLockState(context, event) {
         };
         
         var payload = {
-            lockState: event.payload.lockState == "LOCKED" ? "LOCKED" : "UNLOCKED" //signal success
+            lockState: event.payload.lockState //signal success
         };
         
         var result = {
--- a/oh2.js
+++ b/oh2.js
@@ -86,7 +86,6 @@ exports.handleControl = function (event, context) {
     case 'DecrementPercentageRequest':
         adjustPercentage(context, event);
         break;
-            
     case 'GetLockStateRequest':
         getLockState(context, event);
         break;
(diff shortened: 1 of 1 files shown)


digitaldan:
Hi thanks for the PR!  We are stuck right now on the semantics of how to tag for this, see https://github.com/eclipse/smarthome/issues/1093#issuecomment-287885956  as soon as we figure this out I will review your changes.

fourthwall:
I see. Sorry for being so naive.
I also should mention that this uses a very simple "switch" mapping. I understand there is probably a need for something more complicated to represent failure states etc.

Perhaps a "string" item would be more approrpriate for a lock, to represent various states. I suppose it would depend on how the locks were exposed. 
Although a failiure is easy to signal to Alexa. You simply reverse the requested lock state and she interprets this as an error. But I'm sure you already knew that :)

As for the question of ontology, I don't want to muscle in on what is a discussion between far more qualified developers. But I think the suggestion of a homekit or alexa tag is probably a good one, whereas if neither is present in the tag, it is exposed by both homekit and alexa.

fourthwall:
Thinking about it easiest representation of a lock would be a number. 0 1 2 = unlocked locked failed, or similar. Alexa skill can interpret this when sending back the confirmation to either respond with a matched request (success) or an inverted request (failiure by jamming or otherwise not changing to desired state)

oh2.js
```diff
@@ -469,6 +547,12 @@ function discoverDevices(token, success, failure) {
                   var actions = null;
                   var additionalApplianceDetails = {};
                   switch (tag) {
+                  case "Locks":
```
digitaldan: Can you change this to "Lock" ?  This will conform to the other types.

oh2.js
```diff
@@ -86,6 +86,13 @@ exports.handleControl = function (event, context) {
     case 'DecrementPercentageRequest':
         adjustPercentage(context, event);
         break;
+
```
digitaldan: remove space

oh2.js
```diff
@@ -123,6 +130,77 @@ function turnOnOff(context, event) {
...
+            };
+        };
+        
+        var failiure = function (error) {
+            payload = {
+                lockState: event.payload.lockState == "UNLOCKED" ? "LOCKED" : "UNLOCKED" //reverse to signify error
+            };
+        };
+
+        rest.getItem(event.payload.accessToken, event.payload.appliance.applianceId, success, failiure);
```
digitaldan: Why do you retrieve the state of the item after setting it?

oh2.js
```diff
@@ -123,6 +130,77 @@ function turnOnOff(context, event) {
     rest.postItemCommand(event.payload.accessToken, event.payload.appliance.applianceId, state, success, failure);
 }
 
+function getLockState(context, event) {
```
digitaldan: Can you move the lock functions below the adjustTemperatureWithItems function?  This way we keep the on/off and Percent calls, temp calls, and lock calls organized.

oh2.js
digitaldan: remove white space

digitaldan:
@fourthwall , I think we should just go forward with implementing lock support, I have left a few comments for review.

oh2.js
fourthwall: Hi - this was to see if the change took effect. I feared that if a lock was set and failed to toggle for whatever reason, the item would still be reported as locked when it is not - and according to Amazon's spec, it should return an error if the item failed to lock.

Maybe I did not implement this behaviour correctly.

fourthwall:
I see. I didn't realise there was such a delay.

I'll make the changes.

oh2.js
digitaldan: I think the problem is that no matter what, this will return back the state we just set as a lock will take some time (1 second, 10 seconds, something) and this call will happen milliseconds after we set the state.  Do you agree or do you have other thoughts about how that would work?

fourthwall: Alexa expects an immediate response as to the updated state of the item, so it sounds like your suggestion is the only solution. It may be permissible to add a timeout before verifying the state, but I don't know.

I'd be interested to know what natively supported locks are expected to do in failure condition.

digitaldan: > It may be permissible to add a timeout before verifying the state, but I don't know.

This could be $ costly for us, Lambda functions are charged by execution time, so we need to perform our actions quickly and return.   Still as I think about this, maybe having a 5 or ten second delay makes sense, or better have it poll the item once per second for 5 or ten seconds?  The usage should be fairly low on this feature, so maybe $'s are not a valid concern. 

BTW, I was not correct when I said that the call would return the state we just set, I had a small lapse in thinking :-)  Sending a command to a item will not necessarily set the state of that item, so the call may immediately return the last state ( without some type of wait/sleep/loop) .  Still the issue is the same.

fourthwall: I do not balance the books at the OpenHAB foundation, but I think it's a valid concern to be worried about execution time. Seeing as execution time is a commodity here, the frequent but limited polling would be a good idea, seeing as 90% of the time it will be completed relatively quickly.

digitaldan:
Few more comments, first can you add a more descriptive git comment when you push? "Update oh2.js" for every commit makes it more difficult when reviewing all the changes.  Second can you sign your commits according to https://github.com/openhab/openhab-alexa/blob/master/CONTRIBUTING.md ?  

lastly, I merged in another PR which has put your branch in a conflict state, can you resolve please?

Thanks!

fourthwall:
Yes. Sorry. That was when I was initially playing with it. I'll be more descriptive.

oh2.js
fourthwall: How do you feel the implementation should be done?

digitaldan: I'm going back and forth on this one, my first thought was to poll for 5 seconds using setInterval .... but I have now had second thoughts, if OH successfully accepts the command, I think we should also just return success.  OH lock support is very limited and until it has a more robust impl this IMHO is the best way to go.

fourthwall: The only thing I can think of is if its possible to somehow hand off to another AWS instance and have that respond to the Echo. I don't think this is acutally possible though.

So perhaps as you say the solution is to implement this way.

fourthwall:
Hi @digitaldan 

Have implemented changes as per your suggestion - now always reports success to Alexa if the command is posted successfully to OpenHAB.

digitaldan:
Thanks, I will take a look hopefully this weekend.  BTW, I found this interesting:
https://github.com/alexa/alexa-smarthome-validation/blob/master/nodejs/validation.js#L232

Seems like Amazon expects all responses to return within 7-8 seconds, except Lock messages which can go as long as 60 seconds.  This would seem to support the need to poll (or something) to validate the proper lock state.  Hopefully lock support becomes better in OH where we can eventually do this.

fourthwall:
Good find!

Also, I'm sure you guys are already aware of it but I read a while back that Amazon are offering $100 AWS credits to cover running costs for popular services. This might go some way to mitigating costs for polling?

digitaldan:
I had some time to test this yesterday, it seems to work, however Alexa does not respond that the door is locked as described in the docs.  I wonder if we are returning to quick?  Hopefully i can play more this weekend to test.  Have you tried this on your end yet?

fourthwall:
Hi Dan - I am not located in the United States and as far as I know Amazon only enables various functions for US customers. Although I am moving to the US soon (hopefully) and would probably want to take advantage of the feature so tried to code something out of interest and a willingness to contribute to the project.

It might be worth using the AWS platform itself to craft test cases to send to the lambda code and viewing the response - I did this when playing around with it in python and it gave me the payload dump.

fourthwall:
I might be able to enable the US services by setting my Lambda location to N.Virginia and Alexa locale to US. I'm not sure though.

digitaldan:
Thanks, I'll have to dust off my test cases, I did this early on then just switched to using the echo sim https://echosim.io/

fourthwall:
Neat. 

The easiest thing I did was literally just copy the examples from the Alexa skill reference page into the test case editor on AWS, and looked at the responses.

digitaldan:
So I just tried it on a real echo and it works fine, I'm going to go ahead and merge, thanks!

PR #28, opened 2017-05-12
Log refactor

digitaldan:


digitaldan:
Bad name, its log refactor AND standardizing on single quotes.

PR #29, opened 2017-05-23
Support using alexa: namespace

digitaldan:
Signed-off-by: Dan Cunningham <dan@digitaldan.com>


digitaldan:
as discussed in https://github.com/eclipse/smarthome/issues/1093

PR #42, opened 2017-08-21
Update README.md

2devnull:
Added icons.


2devnull:
Will create new one with original images.

Issue #68, opened 2018-04-28
V3 Metadata and tags


digitaldan:
@jsetton , @fmeies  Right now the V3 skill was built with the idea of tags, as of 2.3 this will change to the Metadata on items, although v2 tags will still be supported (as you both already know, just stating the obvious) .  Right now we have a metadata function in the code that converts metadata to tags, but I wonder if this is the right approach? Now that I have more eyes on this, I would not mind some feedback about how we are translating items and properties to the internal map that we store on Alexa.  Its rather complicated, which bothers me, but I don't know if there's a better approach.

jsetton:
My initial intention when I implemented the metadata support was to limit the amount of changes around the property map core logic, you had implemented, by keeping it based on tags.

However, I thought about directly using the openHAB metadata object as the property map, that we send in the discovery response, and convert all tags to the metadata object format. This basically comes down to what we want the core logic based off.

I guess if at some point, we would want to retire/limit the tag logic, it would make sense to have a metadata-based core. Another point is that we could have the v3 capabilities and parameters support at the metadata level only while still keeping v2 tags support. That would limit the complexity I think, especially with the logic around configuration management, but the usage of the latest openHAB version will be required to use the new enhancements.

In terms of v2 tags, I am not sure if we are using the right term at this point. I feel "alias" or "metadata" label is more appropriate. My goal in my implementation was to have these labels, similar to the ones we have in v2 skill, so that we don't have to specify each time all the different capabilities, especially for light or switch items.

Just for reference, this is how the openHAB metadata object is currently structured.
```
metadata: {
  alexa: {
    value: "<capability#1|label#1,...>"
    config: {
      <key#1>: <value#1>,
      ...
  }
}
```

digitaldan:
>However, I thought about directly using the openHAB metadata object as the property map,

This was my though as well, one issue however is that metadata is linked to a single item, where many items can be linked to a Alexa endpoint,  would an endpoint contain an array of metadata? (have not really thought that out yet) 

>I guess if at some point, we would want to retire/limit the tag logic, it would make sense to have a metadata-based core. Another point is that we could have the v3 capabilities and parameters support at the metadata level only while still keeping v2 tags support. 

This is my intent as well,  we should not encourage future use of tags for Alexa.

> My goal in my implementation was to have these labels, similar to the ones we have in v2 skill, so that we don't have to specify each time all the different capabilities, especially for light or switch items.

Can you give an example of this?

[...]

PR #69, opened 2018-04-30
Metadata Support Updates

jsetton:
Updates related to #68.

* Changed core logic based on item metadata object
* Moved v3 parameters support through tags to metadata only
* Updated concept of metadata labels aligned with v2 tags


digitaldan:
Thanks @jsetton , I will take a look this weekend, was caught up with work all week.

jsetton:
Thanks @digitaldan . Are the changes I made in line with what you expected?

PR #79, opened 2018-07-07
Update USAGE.md

thedannymullen:
This problem has caused many posts on the community forum.  I believe it is because this document sites the items file help as a reference.  In that document it states labels are optional.  However this skill requires them.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -30,6 +30,7 @@ See the [Hue Emulation](https://www.openhab.org/addons/integrations/hueemulation
   Some examples of tagged items are:
   
   ```java
+  Switch Item_Name(Required) "Item Label"(Required) <Icon Name>(optional) [Tag](Required) {channel="..."}(optional)
   Switch Kitchen_Light "Kitchen Light" <light> (gKitchen) ["Lighting"] {channel="..."}
   Dimmer Bedroom_Light "Bedroom Light" <light> (gBedroom) ["Lighting"] {channel="..."}
   Number Bedroom_Temperature "Bedroom Temperature" (gBedroom) ["CurrentTemperature"] {channel="..."}
--- a/USAGE.md
+++ b/USAGE.md
@@ -45,9 +45,9 @@ See the [Hue Emulation](https://www.openhab.org/addons/integrations/hueemulation
     * The interactive REST API interface available through the openHAB 2 dashboard
     * PaperUI does not yet allow manipulation of tags
 
-#### Item Label Recommendation
+#### Item Label Recommendation and Requirement
 
-Matching of voice commands to Items happens based on the Item label (e.g. "Kitchen Light").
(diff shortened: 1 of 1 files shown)


digitaldan:
Hi Danny, I would rather specify that both a tag and a label are required in the same paragraph, so maybe something alter our existing paragraph like so:

`
Items are exposed to Alexa through the use of tags which follow the HomeKit binding tagging syntax. See the Hue Emulation and HomeKit Add-on documentation for details about tagging and available tags. Items also require a label, which are used by Alexa as the spoken name`

thedannymullen:
I understand what you are saying.  Piecing together all of the documents it is not **explicitly** stated that a label is required.  

I updated the docs to now state this.  
Thanks!

USAGE.md
```diff
@@ -30,6 +30,7 @@ See the [Hue Emulation](https://www.openhab.org/addons/integrations/hueemulation
   Some examples of tagged items are:
   
   ```java
+  Switch Item_Name(Required) "Item Label"(Required) <Icon Name>(optional) [Tag](Required) {channel="..."}(optional)
```
digitaldan: So the example here is meant to be a working item file example, the line you added here is not a valid syntax, so I would prefer not to add that here.

USAGE.md
```diff
@@ -45,9 +46,9 @@ See the [Hue Emulation](https://www.openhab.org/addons/integrations/hueemulation
     * The interactive REST API interface available through the openHAB 2 dashboard
     * PaperUI does not yet allow manipulation of tags
 
-#### Item Label Recommendation
+#### Item Label Recommendation and Requirement
```
digitaldan: I don't think this belongs here, this is a section header which summarizes the info proceeding it.

USAGE.md
```diff
@@ -45,9 +46,9 @@ See the [Hue Emulation](https://www.openhab.org/addons/integrations/hueemulation
...
 
-Matching of voice commands to Items happens based on the Item label (e.g. "Kitchen Light").
+Matching of voice commands to Items happens based on the Item label (e.g. "Kitchen Light").  (Note: Although, it is valid syntax to not have an item label, the string in quotes following the name.  For this skill to function properly a label is required.)
```
digitaldan: This seems like an odd place to talk about requirements, can we state labels are required in the section above, where we also talk about needing a tag?

digitaldan:
I would prefer we just update the section I mentioned in my first comment, I think that should suffice.

USAGE.md
```diff
@@ -45,9 +45,9 @@ See the [Hue Emulation](https://www.openhab.org/addons/integrations/hueemulation
...
```
thedannymullen: I am ok with any change that simply sites a label is required.  I seem to not know my way around git very well and lost some of my changes.  

I deleted the one that had a non-working example as you requested.  I kept the other which I think just says a label isneeded.  I just can see the diffs properly to know if it is in the correct place.

thedannymullen:
I am good with the first paragraph.  It just needs to be stated somewhere.   I can update in the morning.

PR #82, opened 2018-08-16
Fix temperature conversion for v3 thermostats

crwatkins:
There seemed to be no attempt at Celsius/Fahrenheit conversions for setting temperatures on v3 thermostats.  This should fix both setTargetTemperature() and adjustTargetTemperature().

### DIFF
--- a/ohConnectorV3.js
+++ b/ohConnectorV3.js
@@ -322,7 +322,18 @@ function setTargetTemperature() {
   Object.keys(properties).forEach(function (propertyName) {
     if (directive.payload[propertyName]) {
       var state = directive.payload[propertyName].value;
+      var scale = directive.payload[propertyName].scale;
       var itemName = properties[propertyName].itemName;
+      var itemScale = properties[propertyName].parameters.scale;
+      // Convert between Celsius and Fahrenheit as needed
+      switch (scale + itemScale) {
+	case "CELSIUSFahrenheit":
+	  state = state * 9 / 5 + 32;
+	  break;
+	case "FAHRENHEITCelsius":
+	  state = (state - 32) * 5 / 9;
+	  break;
+      }
       log.debug("Setting " + itemName + " to " + state);
       promises.push(new Promise(function(resolve, reject) {
         log.debug("PROMISE Setting " + itemName + " to " + state);
@@ -366,9 +377,21 @@ function adjustTargetTemperature() {
   var properties = propertyMap.ThermostatController;
   if (properties.targetSetpoint) {
     var itemName = properties.targetSetpoint.itemName;
+    var itemScale = properties.targetSetpoint.parameters.scale;
     rest.getItem(directive.endpoint.scope.token,
       itemName, function (item) {
-        var state = parseFloat(item.state) + directive.payload.targetSetpointDelta.value;
+        var delta = directive.payload.targetSetpointDelta.value;
+        var scale = directive.payload.targetSetpointDelta.scale;
+        // Convert between Celsius and Fahrenheit delta as needed
+        switch (scale + itemScale) {
+	  case "CELSIUSFahrenheit":
+	    delta = delta * 9 / 5;
+	    break;
+	  case "FAHRENHEITCelsius":
+	    delta = delta * 5 / 9;
+	    break;
+        }
+        var state = parseFloat(item.state) + delta;
         postItemAndReturn(itemName, state);
       }, function (error) {
         context.done(null, generateGenericErrorResponse());


digitaldan:
Thanks for contributing!  I do have a question, In the V3 skill api Amazon now supports reporting the temperature in a device's native format, unlike v2 were we had to convert F to C and back again if a user was using Fahrenheit values.  I'm not sure I  understand the use case for conversion in this PR?  If the item is exposing this as a F value (scale="Fahrenheit"), then why do we need to convert here?

crwatkins:
I have Alexa devices configured for interacting in C and for interacting in F. The payload in the event to the lambda function arrives with a scale of F for the F echos and a scale of C for the C echos.  

If there is a way to request that the payload contain a requested scale, I didn’t know about that.  

When a target temperature arrives in the lambda event in F I need it converted to C to POST because my item is in C.  Did I misunderstand the API?

crwatkins:
@digitaldan, my  [jumbled comment](https://github.com/openhab/openhab-alexa/pull/82#issuecomment-414013209) above was written while I was on the go.  To summarize, I believe:

* The v3 interface allows you to report a scale in a response, so no conversions are necessary in responses

* In a v3 request, temperatures are requested using the scale of the requesting Alexa device, so a conversion **is**  necessary if it does not match the scale (C or F) of the  OpenHAB item.

jsetton:
@crwatkins My understanding is that Alexa thermostat requests include the same scale that is provided in the relevant SetPoint property response. That scale is configurable on the OH side; a thermostat Alexa-enabled group will use Celsius scale by default or Fahrenheit if specified as parameter.

As @digitaldan mentioned, I don't believe there should be any conversion done at the skill level if the OH configuration is setup properly.

Maybe you can include your thermostat items configuration to support your use case.

crwatkins:
@jsetton, 

> My understanding is that Alexa thermostat requests include the same scale that is provided in the relevant SetPoint property response. That scale is configurable on the OH side; a thermostat Alexa-enabled group will use Celsius scale by default or Fahrenheit if specified as parameter.

That's not my experience.  Have you tested with both Celsius configured devices and Fahrenheit configured devices?

An Alexa device that makes a request such as:

```
            "namespace": "Alexa.ThermostatController",
            "name": "SetTargetTemperature",
            "payloadVersion": "3",
```
Will specify the following payload if configured for Celsius:

```        "payload": {
            "targetSetpoint": {
                "value": 23,
                "scale": "CELSIUS"
            }
        }
```
If that same exact Alex device is changed (in Alexa device settings) to not use metric, the request will look like this:

```        "payload": {
            "targetSetpoint": {
                "value": 74,
                "scale": "FAHRENHEIT"
            }
        }
```
> As @digitaldan mentioned, I don't believe there should be any conversion done at the skill level if the OH configuration is setup properly.

Can the scale be somehow tagged in the OH REST API with automatic conversion in OH that I am unaware of?  The existing code doesn't work unless OH is using the same scale as all of the Alexa devices on the account (e.g. OpenHAB "configured" for Celsius (default) and an Alexa device set to Fahrenheit).  How can this be fixed?

jsetton:
> If that same exact Alex device is changed (in Alexa device settings) to not use metric

Are you saying you can specify the scale settings on the Alexa side? I am not aware of that setting. I just took a look at my Alexa app and I couldn't see such setting.

Could you provide your thermostat items configuration in OH? That would be easier to run some test to determine if this is a legit issue or just a configuration issue on your side. Thanks.

crwatkins:
> Are you saying you can specify the scale settings on the Alexa side? I am not aware of that setting. I just took a look at my Alexa app and I couldn't see such setting.

On alexa.amazon.com:  ```Settings -> Devices (name) -> Measurement Units - Temperature Units```
It's similar, but not exactly the same on mobile apps.  Perhaps it could be region specific.

> Could you provide your thermostat items configuration in OH? That would be easier to run some test to determine if this is a legit issue or just a configuration issue on your side. Thanks.
```
/* Dining HVAC */
Group Thermostat_F1_Dining	"Dining Room Thermostat"		<heating>	(dev, F1_Dining) [ "Thermostat", "Fahrenheit" ]
Number HVAC_Setpoint_F1_Dining_F	"Dining Room Thermostat Setpoint [%.1f F]"	<heating>	(Thermostat_F1_Dining, TemperatureSetpoint_F, RestoreOnStartup) [ "TargetTemperature" ]
String HVAC_Mode_F1_Dining	"Dining Room Thermostat Mode [%s]"	<climate>       (Thermostat_F1_Dining, RestoreOnStartup) [ "homekit:HeatingCoolingMode" ] {mqtt=">[broker:home/diningroom/hvac/mode:state:*:MAP(hvac-mode.map)]"}
Number Temperature_F1_Dining_F 	"Dining Room Temperature [%.1f °F]"	<temperature>	(F1_Dining, TemperatureSensor_F, Temperature, Thermostat_F1_Dining, Temperature_Master_Thermostat_F, Chart_Indoor_Temperature) ["CurrentTemperature", "Fahrenheit"]
```

jsetton:
Thanks for the information. I had forgotten about that setting. I was able to replicate the issue. It is indeed related to the device temperature unit setting when it differs with the thermostat scale setting in OH. My initial assumption was incorrect as the scale used in the Alexa request is actually derived from that setting and not from the property response.

However, this is an odd setup. Usually the scale is normalized across system. This is the reason why we can specify the scale. Out of curiosity, any reason why your thermostat item cannot process temperature in Celsius?

Going back to a previous question you had, a way to have OH convert the temperature scale internally is to use the [Unit of Measurement](https://www.openhab.org/docs/concepts/units-of-measurement.html) concept. We could have the skill detect the temperature setpoint item type and if it is of QuantityType, we would send the temperature value and scale via the OH REST API.

Using your item as example, if you set `HVAC_Setpoint_F1_Dining_F` as `Number:Temperature`, the skill would just send `20 °C` to that item when requested via Alexa. And since it is configured as Fahrenheit, it would automatically convert it.

I don't think we should be doing the actually calculation in the skill for just that specific case.

crwatkins:
> However, this is an odd setup. Usually the scale is normalized across system. This is the reason why we can specify the scale. Out of curiosity, any reason why your thermostat item cannot process temperature in Celsius?

Interestingly enough, my HVAC device (which is my code) does handle everything in Celsius and everything in my OH was originally in Celsius.  The issues I had were in UI (e.g. Alexa, HABPanel, basicui, REST) presentation when my reference values were in C and my presentation was in F.  For example, I would experience Alexa speaking degrees out to many decimal places because of rounding (or lack thereof).  Switching my reference to F (as much as I found it disturbing) solved all of these issues at once and is working great now.

jsetton:
Maybe the Unit of Measurement concept I mentioned above could resolve some of the presentation you mentioned.

Going back to my suggestion, I realized that it will not work well with incremental request which would still require the skill to do some conversion. So at this point, I don't any other alternative solution to what you proposed. The only suggestion would be to add a test case for this specific scenario.

jsetton:
@crwatkins So I got to look again at this PR and potentially try to add to the latest update since your change can't be merged as is at this point. But after reading it again, I feel now there is a better way to achieve that.

Two things based on what you sent previously:

>The issues I had were in UI (e.g. Alexa, HABPanel, basicui, REST) presentation when my reference values were in C and my presentation was in F. For example, I would experience Alexa speaking degrees out to many decimal places because of rounding (or lack thereof).

I think once we implement #86 that should address that issue as you would be able to control the number of decimal places.

>Can the scale be somehow tagged in the OH REST API with automatic conversion in OH that I am unaware of?

There is something that can be done with Unit of Measurement concept where we could include the scale in the rest api item update commands for items supporting that capability (e.g. `Number:temperature`). I think that should help leverage OH to do the conversion and with the combination of the other enhancement should resolve your issue.

As I am still somehow confused on how your setup is configured, maybe if you could provide details on how your scale preferences is configured in Alexa and OH along with how you are requesting the temperature commands to Alexa and the response you are expecting. Thanks

crwatkins:
This has been so long ago that I might remember some of the details wrong, but as I describe above, V3 makes requests in both C and F.  The developer has no control over it and must deal with it.  The code at the time 8c7860e did not deal with it.

It doesn't really matter what my setup is as it has changed since I opened this PR (as I described above and likely confused the situation).  Here is the bug report:

Say the following two things to the same Echo:

1.  Set the thermostat to 20 degrees Celsius
2.  Set the thermostat to 68 degrees Fahrenheit

Expected results:  Thermostat is set to the same temperature each time

Observed results from  8c7860e:  One of those commands will set the thermostat wildly wrong
Observed results after this PR 39217c9:  Expected results

Note that this will always fail when there is a mismatch of the units requested by the echo user (either explicitly or by the default of the device) and the units expected by OpenHAB (that was my original situation).

I agree that putting the scale in the REST API using Unit of Measurement would solve this problem.
I totally get that you want to avoid unit conversation in this code.  I needed a solution six months ago; I don't actually have the need for this change anymore, and the patch has a conflict, so I'll just close this PR.

jsetton:
Thanks for your explanation. I have added the conversion in the code. While Unit of Measurement could solve the conversion for setting temperature, it won't for converting delta setpoints.

PR #98, opened 2018-10-31
Fix StepSpeaker discovery

fmeies:
The StepSpeaker interface seems to be missing in the device discovery.


fmeies:
A test for this issue is still missing. I'll close this one and add another pull request.

PR #99, opened 2018-10-31
Fix StepSpeaker discovery

fmeies:
The StepSpeaker interface seems to be missing in the device discovery. Please note that the StepSpeaker interface does not support any reportable properties, see [Alexa.StepSpeaker Interface](https://developer.amazon.com/de/docs/device-apis/alexa-stepspeaker.html#properties)


digitaldan:
Hi, thanks for the commit , i have not tested step speaker yet (obviously), but LGTM.

PR #106, opened 2018-12-05
Added v3 tagging back into the file

Clifford-Foster:
Added DigitalDan's tagging update into the master.  Without this groups in the v3 API aren't properly discovered when using tagging.


digitaldan:
So i don't see how this can be true, `tagsToPropertyMap` is never called anywhere so this has put back in a function that is not used.  I'm going to close this, but will reopen if i missed something.

PR #109, opened 2019-01-05
Update Alexa usage readme

mgeramb:
Explain difference between Alexa Skill and Amazon Echo Control

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -1,6 +1,16 @@
+# Amazon Alexa / Amazon Echo
+
+An intelligent voice guided personal assistant. Amazon Alexa became famous by the Amazon Echo speaker but is also available through independent solutions.
+
+The Alexa (Echo) devices can be integrated in two ways in openHAB:
+
+* Control Items defined in openHAB with your voice through Alexa (e.g. switch on a light defined in openHAB). For this direction, you have to use the Amazon Alexa Skill. See installation instruction below for more information.
+
+* Control Amazon Echo devices from your openHAB (e.g. start/pause Alexa music player, use the Echo device as a Text-To-Speech output device) installation. For this way, install the [Amazon Echo Control Binding](https://www.openhab.org/addons/bindings/amazonechocontrol/).
+
 # Amazon Alexa Skill
 
-An intelligent voice guided personal assistant. Amazon Alexa became famous by the Amazon Echo speaker but is also available through independent solutions. The certified openHAB Alexa skill allows to naturally control the openHAB smart home by voiced commands.  
+The certified openHAB Alexa skill allows to naturally control the openHAB smart home by voiced commands.  
 
 This page describes how to use the [openHAB Alexa Smart Home Skill](https://www.amazon.com/openHAB-Foundation/dp/B01MTY7Z5L).
 The skill connects your openHAB setup through the [myopenHAB.org](http://myopenHAB.org) cloud service to Amazon Alexa.


mgeramb:
Is there anybody who can review this PR?

digitaldan:
hi @mgeramb . Sorry on the delay and thanks for the submission .  I believe the problem you are solving is that there is some confusion on the fact that we have two independent Alexa projects, well actually there are three, we also have the Hue emulation binding which i wrote before Amazon had released the home skill api (makes OH look like a Hue gateway that alexa can control locally) .

I think this makes sense, but i would like to recommend a change to your PR.  I would prefer we start of with something like "This is the Alexa Home skill app for OH"  And then give the description of what it does.  Below that I would have list of "Other Alexa Integrations" something like 

* Other Alexa Integrations
  * [Amazon Echo Control Binding](https://www.openhab.org/addons/bindings/amazonechocontrol/)
     * Control Amazon Echo devices from your openHAB (e.g. start/pause Alexa music player, use the Echo device as a Text-To-Speech output device) installation.
  * [Hue Emulation Binding](https://www.openhab.org/addons/integrations/hueemulation/)
    * exposes openHAB items as Hue devices to other Hue HTTP API compatible applications like an Amazon Echo, Google Home or any Hue compatible application.
 
I want to make sure the language in this doc is focused on our skill to avoid confusion when people com here.

I do this this is important, so thanks for bringing this up!

mgeramb:
hi @digitaldan,
I see the following problem with your suggestion:
If you take a look at the homepage [https://www.openhab.org/docs/ecosystem/alexa/](https://www.openhab.org/docs/ecosystem/alexa/) you will see that this readme is for the user point of view only related to Alexa, not to the Alexa skill. So this is the reason, why the 3 solutions should be described in the same level on this page. I understand, that the file itself is placed in the Alexa-Repository, so that it makes sense, that the other solutions are placed under "Other Alexa Intergration", but this is not shown to the user in this way. What do you think about this? Maybe an overview file should be placed in another repository and links to different readme. But I do not know, how this can be done. 
Best,
Michael

digitaldan:
I see your point now, i was not aware this file was being used as the  Alexa "ecosystem" page.  I think I may need to either then re-write this doc or talk with other maintainers about creating a page different from this one to be used there so we can import the usage doc  to the amazon control binding as well.  Let me talk with some folks and get back to you.  Thanks again for bringing this up.

mgeramb:
In my opinion, the best solution would be to have a new overview page.

mgeramb:
Any progress on this discussion where we can add this information?

jsetton:
Why not just add an item called "Amazon Echo" linking to the addons page and rename the one for this skill to `alexa-skill` under the [doc sidebar](https://github.com/openhab/openhab-docs/blob/master/.vuepress/docs-sidebar.js#L73)?

jsetton:
@mgeramb trying to close this one. What do you think about the change below? It's shorter but I think it allows users to understand the difference between the two features. As far as the link included, I pointed to the section at the bottom of the page but maybe I should just point directly to the binding page. The section at the bottom could be useful for other users that didn't go over the top level section.

https://github.com/jsetton/openhab-alexa/commit/5ae4e637f40a437b6f6d6c562fabf69cd7c4a2ed

mgeramb:
@digitaldan I think you can close this PR for now. Anyway, I think we should create an overview page with all Alexa connections: Skill, amazonechocontrol and HUE emulation. I believe that we need some drawings with the connections and possibilities of the different approaches. But currently I have no time for this...

jsetton:
> Anyway, I think we should create an overview page with all Alexa connections: Skill, amazonechocontrol and HUE emulation. I believe that we need some drawings with the connections and possibilities of the different approaches.

Agreed. Closing this PR for now.

PR #110, opened 2019-01-19
Mention used payload version

FlorianSW:
It took me some time to figure out, that this skill only supports the v2 payload, instead of the preselected and non-legacy v3 version.

Signed-Off-By: Florian Schmidt <florian.schmidt.welzow@t-online.de>


digitaldan:
thanks florian , I'm about to merge v3 into master so hopefully this will never be seen ;-)

PR #138, opened 2019-02-16
Property Response Improvements

jsetton:
These are the improvements that I had originally included in #74. I removed all the code related to the asynchronous deferred response support. I know it's a big change but it includes a bunch of property response improvements along with some code refactoring that would allow us to easily maintain the code going forward. It also includes logic to enforce item type per capability.

I have been running this code as my OH skill for the last 8 months without any issue. I also made sure to include and also improve, in some cases, all the changes you have committed the last few days. So it should be easy to merge.

Feel free to comment on my changes if you have any questions.


digitaldan:
Thanks, this is indeed a huge PR, its going to take me some time to review, but at a quick glance this all seems reasonable refactoring and clean up of the project.  Due to how much there is changed, I will want to get this merged first before we add anything else otherwise I see merge conflicts in our future.

jsetton:
> Due to how much there is changed, I will want to get this merged first before we add anything else otherwise I see merge conflicts in our future.

I very much appreciate that 😄 Let me know if you have any questions. I am with you on getting this merged as soon as possible.

alexaCapabilities.js
```diff
@@ -95,29 +95,33 @@ AlexaCapabilities.prototype.percentageController = function () {
...
+    });
   }
   if (thermostatMode) {
     supported.push({
       "name": "thermostatMode"
-    })
+    });
+  }
+  var configuration = {};
+  if (typeof thermostatMode.parameters.supportedModes === 'string') {
```
digitaldan: this is throwing an error if variables are not defined, something like this to fix:
```javascript
if (thermostatMode && thermostatMode.parameters && typeof thermostatMode.parameters.supportedModes === 'string') {
    configuration.supportedModes = thermostatMode.parameters.supportedModes.split(',').map(mode => mode.trim());
  }
```

jsetton: Thanks. I forgot the fact that a thermostat mode property may not always be defined however no need to check for parameters. If the property exists than it will have a parameter object defined.

https://github.com/jsetton/openhab-alexa/blob/v3-changes/alexaPropertyMap.js#L126

I will make the change in my local repository and wait for other changes that may found.

```js
if (thermostatMode && typeof thermostatMode.parameters.supportedModes === 'string') {
    configuration.supportedModes = thermostatMode.parameters.supportedModes.split(',').map(mode => mode.trim());
}
```

digitaldan:
I'm good with the other changes, if something breaks we can deal with it, but i like the refactoring you did, we should get this merged and move on!  Thanks, awesome work!

jsetton:
Give me a few minutes. I am pushing the changes for the bug you identified and you can merge the changes.  I also going to add a test case for it. Thanks.

jsetton:
Ok we are set. I added the test case and also simplified the last thermostat discovery test case to focus on what we wanted to test since the rest is taken care of by the other cases.

digitaldan:
Awesome, thanks!

PR #141, opened 2019-02-17
Item State Presentation Format Support

### LINKED ISSUE #86: Control precision of number values passed to Alexa TTS service
Number values like temperature provided by thing channels are directly passed to TTS service of Amazon Alexa engine ignoring item configurations like [%.1f %unit%]. Depending on the precision with which the value was passed number values with high precision could end in an output like “The temperature in the children’s room is twenty-three-point-eight-nine-nine-nine-nine-six-one-eight-five-three-zero-two-seven-three-four-three-seven-five degree”. Of course you can try to manipulate number values as desired but especially when dealing with Quantitype values this ends in a nasty task for OH users to manipulate values in rules, please see also my explanation here: [Openhab issue #4054: [Netatmo] Temperature values with many digits after decimal point]( https://github.com/openhab/openhab2-addons/issues/4054#issuecomment-425678209).
Therefore I suggest to apply the formatting pattern which allows users to get exactly the precision they desire. As @wborn explained [here]( https://github.com/openhab/openhab2-addons/issues/4086#issuecomment-426809498), not all item states are expected to have the same precision. E.g. for Energy (kWh) it's sometimes more desirable to have a higher precision.

jsetton:
Fixes #86

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -24,7 +24,11 @@ Matching of voice commands to Items happens based on the Item label (e.g. "Kitch
 
 ### Item Configuration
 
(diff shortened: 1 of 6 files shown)


digitaldan:
Awesome, lgtm.

PR #145, opened 2019-02-17
Move to using request library and native promises for all rest calls

digitaldan:


digitaldan:
@jsetton , this does contain a breaking change, I think it would be simpler to have a single "baseURL" config option then separate proto/host/port/path parameters.  Since this is a major rewrite, I don't feel compelled to try and support both for backwards compatibility .

jsetton:
That's fine by me. You should update [config_sample.js](https://github.com/openhab/openhab-alexa/blob/v3/config_sample.js)

jsetton:
By the way, do we want to update the function to use node.js 8.10? It's now the default version when creating a new lambda function.

digitaldan:
>By the way, do we want to update the function to use node.js 8.10? It's now the default version when creating a new lambda function.

Yes, I have been using 8.10 for a while with my deployments, and we are using it for beta as well.

jsetton:
Do you want to update the travis config and env file in this PR?

PR #147, opened 2019-02-17
V3 use system scale

digitaldan:


digitaldan:
@jsetton would appreciate a review of this, I'm not sure i'm super happy with how this is implemented.  The feature this introduces is the ability to use the user's chosen region as the default setting for temperature, so for the US this is Imperial, everywhere else is Metric. This will save our US users from have to manually configure every temperature related item they have.

see https://github.com/openhab/openhab-core/blob/a37cceab6767313e1b99643739bdc450428dcbde/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/I18nProviderImpl.java#L158 for how this is used in the OH core if you are curious.

digitaldan:
And my IDE did some auto formatting which i know complicates reviewing the actual feature changes to the file

jsetton:
Let me take a look. Funny enough, I was about to submit a related change using UoM item state to determine scale. But I can see your change is taking it to the next level 😄 

```diff
diff --git a/alexaContextProperties.js b/alexaContextProperties.js
index 252fe46..065a8da 100644
--- a/alexaContextProperties.js
+++ b/alexaContextProperties.js
@@ -235,6 +235,18 @@ AlexaContextProperties.prototype.propertiesResponseForInterfaces = function (int
   var self = this;
   var response = [];

+  // function returning thermostat scale based on either property parameters or
+  //    item state using unit of measurement
+  var getThermostatScale = function (property) {
+    if (property.parameters.scale) {
+      return property.parameters.scale.toUpperCase()
+    } else if (property.item.state.endsWith('°F')) {
+      return 'FAHRENHEIT';
+    } else {
+      return 'CELSIUS';
+    }
+  };
+
   interfaceNames.forEach(function (interfaceName) {
     var properties = propertyMap[interfaceName];
     switch (interfaceName) {
@@ -266,20 +278,17 @@ AlexaContextProperties.prototype.propertiesResponseForInterfaces = function (int
       case "ThermostatController": //Group [Thermostat]
         if (properties.targetSetpoint) {
           var item = properties.targetSetpoint.item;
-          var scale = properties.targetSetpoint.parameters.scale ?
-            properties.targetSetpoint.parameters.scale.toUpperCase() : "CELSIUS";
+          var scale = getThermostatScale(properties.targetSetpoint);
           response.push(self.targetSetpointStateProperty(item.state, scale));
         }
         if (properties.upperSetpoint) {
           var item = properties.upperSetpoint.item;
-          var scale = properties.upperSetpoint.parameters.scale ?
-            properties.upperSetpoint.parameters.scale.toUpperCase() : "CELSIUS";
+          var scale = getThermostatScale(properties.upperSetpoint);
           response.push(self.upperSetpointStateProperty(item.state, scale));
         }
         if (properties.lowerSetpoint) {
           var item = properties.lowerSetpoint.item;
-          var scale = properties.lowerSetpoint.parameters.scale ?
-            properties.lowerSetpoint.parameters.scale.toUpperCase() : "CELSIUS";
+          var scale = getThermostatScale(properties.lowerSetpoint);
           response.push(self.lowerSetpointStateProperty(item.state, scale));
         }
         if (properties.thermostatMode) {
@@ -290,8 +299,7 @@ AlexaContextProperties.prototype.propertiesResponseForInterfaces = function (int
         break;
       case "TemperatureSensor":
         var item = properties.temperature.item;
-        var scale = properties.temperature.parameters.scale ?
-          properties.temperature.parameters.scale.toUpperCase() : "CELSIUS";
+        var scale = getThermostatScale(properties.temperature);
         response.push(self.temperatureSensorStateProperty(item.state, scale));
         break;
       case "LockController": //Switch [Lock]

```

jsetton:
So in the same idea than the change I mentioned above, I believe we should be determining the scale in the Alexa context properties. Why not doing the rest call just before calling the `propertiesResponseForInterfaces` and pass its value as an argument of that function?

https://github.com/openhab/openhab-alexa/blob/v3/ohConnectorV3.js#L641

digitaldan:
>Why not doing the location rest call just before calling `propertiesResponseForInterfaces`

This would need to happen to every report state and command call that involves temperature, correct?   That seems like a lot of redundant calls which add a little more delay .  Also this value will probably never, ever change once a user installs OH, so retrieving it every time seems unnecessary.  I would prefer to grab it once and set it as a global property, unfortunately the cookie/property map is the only place to store this.  

Maybe we should add a new global section to the property map to store such information, I can see that being very useful indeed. Maybe a "OpenHAB" namespace like
```javascript
OpenHAB: {
      globals: {
         scale: "FAHRENHEIT"
       }
  }
```

jsetton:
Yeah that was the other solution I was thinking. I would store the actual system value and determine the scale value when needed.

Although I am just wondering what other properties would be benefiting from that global setting. It seems a lot of overhead especially with the extra call while we basically have that information in the item state using UoM, which basically derives from that global setting.

I understand that it still requires a configuration change from the users perspective opposed to just getting the information but I would expect most user switching to UoM-enabled items and if they don't, they would still have the ability to use parameters.

digitaldan:
>I just did a quick test and was able to confirm that with my OH setup, I set a temperature item with no state presentation format and the default UoM temperature scale is set to F.

I'm ok with this then, makes things simpler for sure, we should probably make it clear in the usage dock that temperature items should use UOM.  

Would it also be wise to try the state value first, and if no C or F, then stateDescription second (for c or f)  ? I think that would probably cover 99% of all use cases, and in my own setup would work 100% across my items, new and old.  This is also what OH does according the UOM docs.

Thanks for the feedback.

digitaldan:
and by stateDescription i mean stateDescription.pattern

jsetton:
> I'm ok with this then, makes things simpler for sure, we should probably make it clear in the usage dock that temperature items should use UOM.

I agree but it shouldn't be a requirement. Amway I think most of the recently updated 2.x thermostat control bindings are already leveraging that feature.

> Would it also be wise to try the state value first, and if no C or F, then stateDescription second (for c or f) ? I think that would probably cover 99% of all use cases, and in my own setup would work 100% across my items, new and old. This is also what OH does according the UOM docs.

I am not 100% sure to follow you. My understanding is that UoM item raw states will always be formatted as `<value> <unit>`. It is not affected by the item state presentation pattern. So we don't need to look at that setting in this case, unless I am missing something. In terms of precedence, I would assume that we would put Alexa metadata parameter over UoM unit?

So looking at PR #146 I submitted, we will actually need to have the scale available in the property map if we don't want to default to "CELSIUS" during the conversion if not defined. I am currently making some changes to account for this. Give me a few minutes and I will push something to my repo so you can let me know what you think.

jsetton:
What do you think about this change? It's very simple and takes care of what we need and we can increase with additional units if necessary.

```diff
diff --git a/alexaPropertyMap.js b/alexaPropertyMap.js
index f08783d..9042261 100644
--- a/alexaPropertyMap.js
+++ b/alexaPropertyMap.js
@@ -16,6 +16,11 @@ var utils = require('./utils.js');
 var CAPABILITY_PATTERN = /^(?:Alexa\.)?(\w+)\.(\w+)$/;
 // Define item property metadata parameter format pattern
 var ITEM_PARAM_PATTERN = /^item(\w)(\w+)$/;
+ // Define unit of measurement scale mapping
+var UNIT_MEASUREMENT_SCALE_MAPPING = {
+  '°C': 'CELSIUS',
+  '°F': 'FAHRENHEIT'
+}

  /**
  * Defines property map object to assoicate items to an endpoint from metadata, per the description below:
@@ -135,6 +140,12 @@ AlexaPropertyMap.prototype.addItem = function(item) {
         return;
       }

+      // Set scale parameter based on unit of measurement number item type if not already defined
+      if (item.type.startsWith('Number:') && !property.parameters.scale) {
+        var unit = item.state.split(' ').pop();
+        property.parameters.scale = UNIT_MEASUREMENT_SCALE_MAPPING[unit];
+      }
+
        // Extract item property settings from metadata config. These settings starts with 'item'.
       Object.keys(property.parameters).forEach(function(parameter) {
         if (matches = parameter.match(ITEM_PARAM_PATTERN)) {
```

digitaldan:
>I am not 100% sure to follow you.

If a user is not using a Number:Temperature format in their items, then the state returned is just the number value, but its very likely that the stateDescription.pattern will have the F or C in it (legacy stuff) , either set by the user or set by the bindings thing/channel description , .  This would cover users with older style item files.  I'm ok either way, but if it was me, I would cover both bases and make it "just work" in most cases.   I was guilty of this in my own items.  Is that clear, or I have I confused something?

digitaldan:
LGTM
>if (item.type.startsWith('Number:')

Any reason not to match "Number:Temperature" , just curious, not a big deal.

jsetton:
Yeah I got it now. I didn't think about that case as my thermostat items are on the UoM band wagon 😄 The issue I have with looking at the `stateDescription.pattern` is that it can be set to anything by the user although most likely would match what we would expect.

In terms of support, I let you decide if you want to add it or not. For non UoM, I would just document the option to change item to use that feature (I don't even believe that the thermostat control binding needs to support it), or just use standard Alexa metadata parameter configuration.

jsetton:
> Any reason not to match "Number:Temperature" , just curious, not a big deal.

I made it more generic if we wanted to add more unit from other dimensions later on. But I could set it to match `Number:Temperature` only if you want. The idea was to just add new mappings going forward.

Just to be on the same page, the Alexa metadata scale parameter should take precedence over UoM unit mapping? This is how it is represented in my change above.

digitaldan:
>n terms of support, I let you decide if you want to add it or not.

I'm good with this, if it becomes an issue, we have an easy fix, but pushing people to UoM is probably a good thing.

jsetton:
> Just to be on the same page, the Alexa metadata scale parameter should take precedence over UoM unit mapping? This is how it is represented in my change above.

What about this? I just want to make sure we are on the same page. I am actually 50/50 as UoM should be more accurate than a user parameter configuration 😄

digitaldan:
Yep , I think the Alexa tags should take precedence since it has to be explicitly set by the user for just this meta data namespace

jsetton:
Fine by me. The PR I submitted is setup that way.

PR #149, opened 2019-02-18
Contact and Motion Sensors Support

### LINKED ISSUE #95: Support for new contact and motion sensor API
It seems that amazon just activated support for contact and motion sensors that can be used in routines. The primary purpose is to let iot devices trigger rules but i think it should also be possible to let openhab send arbitrary status updates to alexa, which in turn can be used to trigger routines. 

Essentially this means that two-way communication would be possible without the echo-binding which uses an inofficial api and therefore has some authentication issues (at least when using 2FA).

jsetton:
* added generic property state normalization function returning relevant alexa state
* updated item state normalization to only occur when pattern and state defined
* updated display categories list

Fixes #95

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -260,6 +260,18 @@ In openHAB a thermostat is modeled as many different items, typically there are 
     * Supported item type:
       * Player
     * Default category: OTHER
+  * `ContactSensor.detectionState`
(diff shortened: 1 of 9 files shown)


digitaldan:
👍

Issue #123, opened 2019-02-10
[v3] Beta


digitaldan:
We now have the V3 version of the sill deployed in a "Beta Test" through Amazon.  This will allow us to have a small group of users upgrade early to the new skill and provide feedback.

I want to use this issue to coordinate updates and communicate to users who have access to the beta.

When opening issues against the beta, please prefix "[v3-beta]" in the issue title.  Include as much information about:

What you were saying or doing, include any item names or labels that could help in debugging
What the expected result was
What the actual result was
Exact time, including seconds if possible of the issue
What time zone you are in 
What country you are in (to identify which global lambda you are using)

For our first wave of users I want to focus on those using V2 style tags to ensure we have backwards compatibility and that we don't break it for the thousands using the current skill.

Configuration and usage can be found in at https://github.com/openhab/openhab-alexa/blob/v3/USAGE.md

SamuelBrucksch:
Hi, i want to test. How can i participate?

codeaholics:
Yes please!

PheiPheiPhei:
I'd love to participate. I'm in Germany, if that makes any difference.

mgeramb:
I am also very intested joining the beta test!

Flole998:
Same here, I'd also love to join!

digitaldan:
Thanks everyone!!! Can you send me your email (and github username in the email) to  apps@openhabfoundation.org  ?

digitaldan:
As I mentioned in the description, would be great if people joining the Beta would test their existing V2 configuration first, and if that works, then move to using the V3 syntax and new features.

codeaholics:
I've installed the new skill and it found all my devices just fine. I particularly like that Alexa can now query the status (in fact, this is the thing I was most looking forward to with the V3 skill). Seems to work well. I'll keep playing with it for a while.

I haven't tried the new tagging yet.

digitaldan:
@codeaholics so you were using the v2 item tags and everything is working as expected?

codeaholics:
Correct. That said, I don’t have loads of devices right now. Mostly switches configured as “Lighting” and dimmers. I’ll likely be getting Hue soon for some lights in our conservatory. Considering setting them up via OH, so that would allow me to test colours.

codeaholics:
Having said that... I can see that lights are on or off in the Alexa app now, but when I ask her "is [...] on?" she says "[...] doesn't support that".

Flole998:
[...]

PR #153, opened 2019-02-20
Fixes powerstate issue

digitaldan:

### DIFF
--- a/alexaContextProperties.js
+++ b/alexaContextProperties.js
@@ -24,8 +24,8 @@ AlexaContextProperties.prototype.powerStateProperty = function (state) {
   if (state.split(',').length === 3) {
     state = state.split(',').pop();
   }
-  if (!isNaN(state)) {
-    var num = parseInt(state);
+  var num = parseInt(state);
+  if (!isNaN(num)) { 
     state = num > 0 ? 'ON' : 'OFF'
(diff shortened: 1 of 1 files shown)


digitaldan:
@jsetton this seems to fix the issue over at #123 about ON/OFF not working with  dimmers.  Not sure if there are other places we need to do this.

jsetton:
So the issue is related to the item state normalization function. It now adds the state presentation format on these items and the isNaN function isn't able to check properly with a unit. Let me take a look at it further.

digitaldan:
Parseimt and parsefloat both read a string to the end of it or to the first non numeric character , and rerurn nan if no number is found , apparently isnan does not do this same parsing

jsetton:
I am making a change to limit that normalization to only use the first format specifier from the description pattern since the main purpose for it was to control number precision. I don't see why we should open it to such a huge amount of possibilities as it will could bring a lot of unforeseen issues.

Another potential issue with using the full pattern that I found is that the parse functions can't get a number if there is a non numerical character at the beginning.

```
> a = ">40%<"
'>40%<'
> parseInt(a)
NaN
```

jsetton:
I just submitted #154. Let me know if this fix works for you.

PR #154, opened 2019-02-21
Limit item state normalization to only process numerical format placeholders

jsetton:


digitaldan:
Hi @jsetton with this PR is #153 still needed then?

PR #156, opened 2019-02-22
V3 logging fixes

digitaldan:
@jsetton  logging was not printing the second argument in the log function for me,  i ran across this bug https://github.com/winstonjs/winston/issues/1217  on the winston repo that maybe explains why, can you confirm my changes work for you before i merge?

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ var ohv3 = require('./ohConnectorV3.js');
  * Incoming events from Alexa Lighting APIs are processed via this method.
  */
 exports.handler = function (event, context, callback) {
-  log.debug('Input: ' + JSON.stringify(event));
+  log.debug('Input: ', event);
   var version = parseInt(event.directive ? event.directive.header.payloadVersion : event.header.payloadVersion);
   switch (version) {
     case 3:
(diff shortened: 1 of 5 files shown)


jsetton:
I am looking into this right now. FYI, I have another big code refactor improvement PR coming your way, in a bit, around Alexa capability settings and property state management. I think you will like it. 😄

digitaldan:
👍 Awesome, I'm going to open a Bug on something I have a question on , look for it in just a min.

jsetton:
Change looks good. Another option would have been to reintroduce the concatenation of the JSON stringify output to the log message string.

digitaldan:
I'll merge, i kinda like the simplify of it.

jsetton:
So I found another issue with the logger while doing some test. When the provided object includes a `message` property, it uses that property as the message. I am going to rename these properties to `reason` instead. I also confirmed that it is only when that property is at the root level. So none of the Alexa response object should be affected.

Before
`error: getPropertiesResponseAndReturn failed with error: {"message":"Invalid item state returned by openHAB","items":[{"name":"switch1","state":"NULL","type":"Switch"}]}`

Now
`error: Invalid item state returned by openHAB {"items":[{"name":"switch1","state":"NULL","type":"Switch"}]}`

Issue #142, opened 2019-02-17
[V3] Mode, Range, and Toggle  support


digitaldan:
These are US only atm, but looks very cool

https://developer.amazon.com/blogs/alexa/post/2ccbb04d-47c0-41c8-92fe-feb581deb676/understanding-the-updated-smart-home-skill-api-preview

jsetton:
Based on the examples provided in the API documentation, here is my proposed sketch for these interfaces:


[Mode Controller](https://developer.amazon.com/docs/device-apis/alexa-modecontroller.html#example-discovery-response)
```js
Group gWasher "Washer" {alexa="Endpoint.Other"}
Number WasherCycle "Washer Cycle" (gWasher) {alexa="ModeController.mode" [Normal=0,Delicates=1,supportedModes="Normal:Cottons,Delicates:Knites",friendlyNames="Wash Cycle,Wash Setting",ordered=false,local="en-US"]}
Number WasherTemperature "Washer Temperature" (gWasher) {alexa="ModeController.mode" [Cold=0,Warm=1,Hot=2,supportedModes="Cold:Cool,Warm,Hot",friendlyNames="Wash Temperature,assetId:Setting.WaterTemperature",ordered=true,local="en-US"]}
```

[Range Controller](https://developer.amazon.com/docs/device-apis/alexa-rangecontroller.html#example-discovery-response)
```js
Number BasementFan "Basement Fan" {alexa="RangeController.rangeValue" [supportedRange="1:10:1",presets:10="assetId:Value.Maximum,assetId:Value.High,Highest",friendlyNames"assetId:Setting.FanSpeed",locale="en-US"]}
```

[Toggle Controller](https://developer.amazon.com/docs/device-apis/alexa-togglecontroller.html#discovery-response)
```js
Switch TowerFan "Tower Fan" {alexa="ToggleController.toggleState" [friendlyNames="assetId:Setting.Oscillate,Rotate",local="en-US"]}
```

These are indeed very powerful. Regarding the [asset ids](https://developer.amazon.com/docs/device-apis/resources-and-assets.html#global-alexa-catalog), this is a new feature part of the capability ressources included in these interfaces.

digitaldan:
Thanks @jsetton I like it!  i think once you read the api docs it make sense, for users we will need to have lots of examples  and a tables of assetIds  in our usage doc to clearly explain.

PR #162, opened 2019-03-05
Mode Range Toggle Controller Support

jsetton:
Fixes: #142 

**Changes**
* added multi-instance support to accomodate for the new capabilities
* improved test suite to check for new standard capabilities parameters
* revamped unit of measurement support
* removed power level unecessary normalize function
* added kelvin temperature conversation normalization

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -73,7 +73,7 @@ NOTE: the Alexa skill has 3 different percentage interfaces, BrightnessControlle
 
 #### Group Items
 
-While single mapping items works for many use cases, occasionally multiple openHAB items need to be mapped to a single endpoint in Alex.
+While single mapping items works for many use cases, occasionally multiple openHAB items need to be mapped to a single endpoint in Alexa. When using a group item, keep in mind that there can only be one specific interface capability per group. If you need to have more than one instance of a given capability, you should use the Mode, Range and Toggle controllers that are described at the end of this section.
 
 For this example we will use 2 different use cases, a thermostat and a stereo.
 
@@ -121,9 +121,9 @@ In openHAB a thermostat is modeled as many different items, typically there are 
   ```
 ```
   Group Fan     "Fan"          {alexa="Endpoint.Other"}
-  Number Speed  "Speed"  (Fan) {alexa="RangeController.rangeValue" [supportedRange="1:10:1",friendlyNames="Setting.FanSpeed,Speed",preset-1="Value.Minimum,Value.Low,Lowest",preset-10="Value.Maximum,Value.High,Highest"]}
+  Number Speed  "Speed"  (Fan) {alexa="RangeController.rangeValue" [supportedRange="1:10:1",presets="1:Value.Minimum:Value.Low:Lowest,10:Value.Maximum:Value.High:Highest",friendlyNames="Setting.FanSpeed,Speed"]}
   Switch Rotate "Rotate" (Fan) {alexa="ToggleController.toggleState" [friendlyNames="Setting.Oscillate,Rotate"]}
-  Switch Power  "Power"       (Washer) {alexa="ToggleController.toggleState" [friendlyNames="DeviceName.Fan"]}
+  Switch Power  "Power"  (Fan) {alexa="ToggleController.toggleState" [friendlyNames="DeviceName.Fan"]}
   ```
 
 #### Supported item mapping metadata
@@ -163,7 +163,6 @@ In openHAB a thermostat is modeled as many different items, typically there are 
         * Celsius
         * Fahrenheit
         * Kelvin
-      * defaults to scale=Celsius if omitted.
   * `ThermostatController.upperSetpoint`
     * Items that represent a upper or HEAT set point for a thermostat, value may be in Celsius or Fahrenheit depending on how the item is configured (e.g., scale=Fahrenheit). If omitted, the scale will be determined based on: (1) unit of measurement unit if Number:Temperature item type; (2) your openHAB server regional measurement system or region settings (US=Fahrenheit; SI=Celsius); (3) defaults to Celsius.
     * Supported item type:
@@ -173,6 +172,7 @@ In openHAB a thermostat is modeled as many different items, typically there are 
       * scale=`<scale>`
         * Celsius
         * Fahrenheit
+        * Kelvin
       * comfortRange=`<number>`
         * When dual setpoints (upper,lower) are used this is the amount over the requested temperature when requesting Alexa to set or adjust the current temperature.  Defaults to comfortRange=1 if using Fahrenheit and comfortRange=.5 if using Celsius. Ignored if a targetSetpoint is included in the thermostat group.
   * `ThermostatController.lowerSetpoint`
@@ -184,6 +184,7 @@ In openHAB a thermostat is modeled as many different items, typically there are 
       * scale=`<scale>`
         * Celsius
         * Fahrenheit
+        * Kelvin
       * comfortRange=`<number>`
         * When dual setpoints (upper,lower) are used this is the amount under the requested temperature when requesting Alexa to set or adjust the current temperature.  Defaults to comfortRange=1 if using Fahrenheit and comfortRange=.5 if using Celsius.  Ignored if a targetSetpoint is included in the thermostat group.
   * `ThermostatController.thermostatMode`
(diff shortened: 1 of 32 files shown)


digitaldan:
Very cool, I will take a look today.

jsetton:
Thanks. Let me know what you think about the last changes [[1]](https://github.com/openhab/openhab-alexa/pull/162/commits/bd8141d2ba62bcd4f4065371b9fce28d5eb70c36) and [[2]](https://github.com/openhab/openhab-alexa/pull/162/commits/ed4ea4e8c0dcf2b32051d5f06f4a6ce9cefa1104), I just pushed removing the `assetId` prefix requirement and combining presets parameters configuration into one.

Basically, for the first change, the prefix prevented the ability to specify asset ids in the `supportedModes` parameter due to the setting format. Since they have a distinct format, it is easy to differentiate them from text. Although, we could do a simple split check prior so that invalid asset don't get reported as just a label text. Or another solution would be to use the full asset names with `Alexa.` prefix. I am open to make either change.

The second change is just bringing supported presets configuration in line with supported modes.

```js
Group Washer       "Washer"               {alexa="Endpoint.Other"}
Number Cycle       "Cycle"       (Washer) {alexa="ModeController.mode" [Normal=0,Delicate=1,supportedModes="Normal:Cottons,Value.Delicate:Knites",friendlyNames="Wash Cycle,Wash Setting",ordered=false]}
Number Temperature "Temperature" (Washer) {alexa="ModeController.mode" [Cold=0,Warm=1,Hot=2,supportedModes="Cold:Cool,Warm,Hot",friendlyNames="Wash Temperature,Setting.WaterTemperature",ordered=true]}  
Switch Power       "Power"       (Washer) {alexa="ToggleController.toggleState" [friendlyNames="DeviceName.Washer"]}
```

```js
Group Fan     "Fan"          {alexa="Endpoint.Other"}
Number Speed  "Speed"  (Fan) {alexa="RangeController.rangeValue" [supportedRange="1:10:1",supportedPresets="1:Value.Minimum:Value.Low:Lowest,10:Value.Maximum:Value.High:Highest",friendlyNames="Setting.FanSpeed,Speed"]}
Switch Rotate "Rotate" (Fan) {alexa="ToggleController.toggleState" [friendlyNames="Setting.Oscillate,Rotate"]}
Switch Power  "Power"  (Fan) {alexa="ToggleController.toggleState" [friendlyNames="DeviceName.Fan"]}
```

digitaldan:
Should have some time today to take a look.

jsetton:
Thanks. I have the SecurityPanelController interface pretty much done as well and was waiting to have this one fully reviewed and merged before sending it your way 😄 

I am keen on hearing your opinion on the configuration format question I asked above. I am trying to get your perspective on the fact that this could potentially be confusing to the users.

digitaldan:
Question, this part if the item:
`"Normal:Cottons,Value.Delicate:Knites"`  
what is the difference between `Normal` and `Value.Delicate`,  What does the 'Value' part mean?

jsetton:
`Normal` is a text label while `Value.Delicate` is an asset id label. It used to be prefixed `assetId:Value.Delicate` in my original sketch but I dropped that prefix in order to support these type of label in `supportedModes` parameter.

```
{
  "@type": "text",
  "value": {
    "text": "Normal",
    "locale": "en-US"
  }
}
{
  "@type": "asset",
  "value": {
    "assetId": "Alexa.Value.Delicate"
  }
}
```

The fact you are asked about it already makes me believe this is confusing. So we will probably need to find another way to name these labels.

digitaldan:
You beat me to it, i was about to say i went and actually read the USAGE.md file and figured it out.  It makes sense once you do that, but i keep thinking about our users first look at everything, its going to be a bit overwhelming .  Not sure if there's anyway around it, other then lots of working examples.

jsetton:
One of the alternative I proposed above was to use the full asset id.
`supportedModes="Normal:Cottons,Alexa.Value.Delicate:Knites"`

jsetton:
Or maybe something like this?
`supportedModes="Normal|Cottons,assetId:Value.Delicate|Knites"`

digitaldan:
So what confused me was that the asset id was both used as a key, but i think also implies an alternative name. So in this example "Knites" is being mapped to the "Delicate" mode, but by using an asset id as the key, you are also mapping "Delicates" and "Delicate" implicitly.  Is this correct?

digitaldan:
what about 

`supportedModes="Mode1:altName1:altName2:altNameN, Mode2:altName1:altName2:altNameN,"`

where altName is either a spoken name or a assetId?

`supportedModes="Normal:Cottons,Delicate:Value.Delicate:Knites"`

jsetton:
Yeah I was about to propose that. It's actually how it is done for RangeController presets. I was just concerned about the item state description integration but can easily be fixed by using `<option.value>:<option.label>` which is actually much better compared to what I am currently doing (using label as state and adding label>value mapping to parameters)

What about making assetId more obviously? I am 100% sure that leaving `Value.Delicates` is self-explanatory.

jsetton:
Actually in the example you provided. It would be more like:
`supportedModes="Normal:Normal:Cottons,Delicate:Value.Delicate:Knites"`

Now I remember why I was reluctant to go that route 😄

digitaldan:
>Now I remember why I was reluctant to go that route

Ah , i see the redundancy in it.  I need to think on it a bit, but i would rather error on the side of clarity than efficiency. If we used the example:  

```perl
Number Cycle       "Cycle"       (Washer) {alexa="ModeController.mode" [NORMAL=0,DELICATE=1,supportedModes="NORMAL:Normal:Cottons,DELICATE:Value.Delicate:Knites",friendlyNames="Wash Cycle,Wash Setting",ordered=false]}
```
I think this would be more clear, wdyt?  Or did i miss something ?

jsetton:
> i would rather error on the side of clarity than efficiency

I totally agree. Regarding your example, I don't think enforcing uppercase state is the best thing. The whole purpose with these interfaces is the high level of customization. What about this? `>` delimits state from friendly names and `@` defines an asset ids like an alias?

`supportedModes="Normal>Normal:Cottons,Delicate>@Value.Delicate:Knites"`

jsetton:
@digitaldan I just updated my proposed change above. What do you think?

digitaldan:
> I don't think enforcing uppercase state is the best thing.

Yes, i agree, i was confusing again that the key was both a spoken value and the mode value.  Had to read the API docs again last night to realize that.

> just updated my proposed change above. What do you think?

Yes, this looks good!  Does this mean we would drop the '[Normal=0,Delicate=1]`  style mode mappings? 

>Shortened format w/ string mapping

I totally get this as well, the only thing i worry about is having more then one way to do something.  Users are going to have little bit of a shock when they see everything they can do on the usage page.  I fear having to explain multiple ways to do the same thing using a custom syntax.  i'm on the fence and can go either way.

jsetton:
> Yes, this looks good! Does this mean we would drop the '[Normal=0,Delicate=1]` style mode mappings?

Correct. I initially thought that the values needed to be unique but I did some testing and it works very well. And actually makes more sense that way. Alexa just sends the OH state directly in the directive.

```
{
    "directive": {
        "header": {
            "namespace": "Alexa.ModeController",
            "name": "SetMode",
            "payloadVersion": "3",
            "instance": "Cycle",
        },
        "endpoint": {
            "endpointId": "Washer",
            "cookie": {
                "propertyMap": "<propertyMap>"
            }
        },
        "payload": {
            "mode": "0"
        }
    }
}
```

> I totally get this as well, the only thing i worry about is having more then one way to do something. Users are going to have little bit of a shock when they see everything they can do on the usage page. I fear having to explain multiple ways to do the same thing using a custom syntax. i'm on the fence and can go either way.

I was thinking about it as well. We could just leave it undocumented for now 😄 Or maybe we could just offer the option that for states by themselves only. Either way, I might just implement as I still think it is a natural way of setting up.

`supportedModes="Normal=Normal:Cottons,Delicate=@Value.Delicate:Knites"`
`supportedModes="Cold=Cold:Cool,Warm,Hot"`

digitaldan:
I like it, ready to merge?

jsetton:
I still need to do the change. I should have it available later today.

jsetton:
I just pushed the change. Let me know if you are fine with the updated usage page. I buried the shortened information in the ModeController description details and provided the standard format in the group endpoint examples at the top of the page. I can remove it if necessary.

digitaldan:
LGTM. Ready to merge?

jsetton:
Yep. I am going to PR the security panel support changes once this PR is merged.

digitaldan:
👍

Issue #111, opened 2019-01-20
[V3] Add security panel controller support


smhgit:
[SecurityPanelController](https://developer.amazon.com/docs/device-apis/alexa-securitypanelcontroller.html)

digitaldan:
Agreed, that would be cool

digitaldan:
So i have started thinking about this one, my thoughts so far as sample items would be:

```
Group  TestAlarm "Alarm" {alexa="SecurityPanelController"}
String TestAlarmModeCommand "Alarm"  (TestAlarm) { alexa="SecurityPanelController.command" [ARMED_AWAY="away",ARMED_STAY="stay",ARMED_NIGHT="night",DISARMED="disarm",pinSupport=true,supportsArmInstant=true] }
String TestAlarmModeState "Alarm"  (TestAlarm) { alexa="SecurityPanelController.armState" [ARMED_AWAY="away",ARMED_STAY="stay",ARMED_NIGHT="night",DISARMED="disarm"] }
Switch TestAlarmBurglaryAlarm "Burglary in Alarm" (TestAlarm) {alexa="SecurityPanelController.burglaryAlarm"}
Switch TestAlarmFireAlarm "FIre in Alarm" (TestAlarm) {alexa="SecurityPanelController.fireyAlarm"}
Switch TestAlarmCarbonMonoxideAlarm "Carbon Monoxide in Alarm" (TestAlarm) {alexa="SecurityPanelController.carbonMonoxideAlarm"}
Switch TestAlarmWwaterAlarm "Water in Alarm" (TestAlarm) {alexa="SecurityPanelController.waterAlarm"}
Number TestAlarmExitDelayInSeconds "Exit Delay In Seconds" (TestAlarm) {alexa="SecurityPanelController.exitDelayInSeconds"}
```

@jsetton didn't know if you had started working on this, I have a rough sketch out of the capabilities and properties, but not much more at this point. 

 For pin support when sending a command, I was thinking we could append the pin to the command if enabled,  so when arming we would send "away:1234" as a string.  If there is a better idea, i'm all ears :-)

jsetton:
Since you add this one assigned to yourself, I left that one last 😄 I was going to aim to range, toggle and time hold capabilities next.

Your sketch looks good. A couple things to note:

[...]

PR #166, opened 2019-03-09
Security Panel Controller Support

jsetton:
Fixes: #111 

**Changes**
* Added ability to restrict the properties included in response context
* Added ability to trigger an error reponse based on properties response name and value
* Moved supported property states into schema config object
* Fixed object not properly cloned causing some reference objects to be unecessarily updated

### DIFF
--- a/alexa/directive.js
+++ b/alexa/directive.js
@@ -71,13 +71,11 @@ class AlexaDirective extends AlexaResponse {
    *
    *
    * @param {Array}  items
-   * @param {Object} parameters     Additional parameters [header, payload, response] (optional)
+   * @param {Object} parameters     Additional parameters [header, payload, properties, response] (optional)
    */
   postItemsAndReturn(items, parameters = {}) {
-    const promises = [];
-    items.forEach((item) => {
-      promises.push(rest.postItemCommand(this.directive.endpoint.scope.token, item.name, item.state));
-    });
+    const promises = items.map(item =>
+      rest.postItemCommand(this.directive.endpoint.scope.token, item.name, item.state));
     Promise.all(promises).then(() => {
       if (parameters.response) {
         log.debug('postItemsAndReturn done with response:', parameters.response);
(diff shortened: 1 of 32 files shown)


digitaldan:
Sweet, i will give it a try today with my DSC panel .

alexa/directive.js
```diff
@@ -105,42 +103,50 @@ class AlexaDirective extends AlexaResponse {
...
           this.propertyMap[capability.interface][capability.property].item = result;
         });
         return result;
-      }));
-    });
+      })
+    ), []);
     Promise.all(promises).then((items) => {
       // Throw error if one of the state item is set to 'NULL'
       if (items.find(item => item.state === 'NULL')) {
```
digitaldan: So i know this is not part of this commit, but this line just bit me.  I had several items in a group, one of them had a NULL value,  but it should not have mattered as the action i was taking (disarm) did not rely on it, but the response failed anyways.  This took me a while to figure out.  Is this line needed ?  I commented it out and it's working now, although i have not tried a command that depends the the item with a null state.

jsetton: I actually had similar issues during my testing and was considering changing the logic but the problem with NULL value is that they can be sent back in the properties response since this is considered as a string value, or could end up being normalized to undefined and then the properties check below would also throw an error.

I am open for suggestions on how to handle this case. The goal obviously is not to send invalid values back. One solution I was thinking was to remove the invalid values (NaN number type, NULL string type or undefined type) from the response and just send the ones that are valid.

Also, my guess why you were getting that error was probably because you have set some of alarm properties in the security panel group and these are getting pulled for the properties response. The logs should show the list of items with the ones that were set to NULL.

digitaldan: >Also, my guess why you were getting that error was probably because you have set some of alarm properties 

Correct, I had NULL value for the state of a "SecurityPanelController.burglaryAlarm" which was causing it grief.  Since the is not used for a arm/disarm response it would be good to not cause it to fail.

I worry about users trying to debug this kind of stuff.  I was able to go through the logs and see debug info, the vast majority of our users will not be able to do this.  It would be nice if the binding was a forgiving as possible

Also I have not played around with how Alexa responds with the different types of error code  here https://developer.amazon.com/docs/device-apis/alexa-errorresponse.html#error-type-values  but it might be nice to make use of some of these and document them in the readme if Alexa can respond differently and we can map that to certain conditions.

jsetton: > Also I have not played around with how Alexa responds with the different types of error code here https://developer.amazon.com/docs/device-apis/alexa-errorresponse.html#error-type-values but it might be nice to make use of some of these and document them in the readme if Alexa can respond differently and we can map that to certain conditions.

So I just pushed a change to this PR to improve error handling. That should cover most of the issues you saw. I also included a [troubleshooting guide](https://github.com/openhab/openhab-alexa/blob/3de977a7b98d1d933643f5b1373c3481f5cd3d67/USAGE.md#troubleshooting).

Let me know what you think. Maybe we should put it higher in the page? Basically, the properties for undefined `NULL` or uninitialized `UNDEF` item states aren't included in the response but will change the health connectivity status to `UNREACHABLE`. Also, when restricting the list of properties to include in response (e.g. arm/disarm requests), the item state for the other properties part of the same interface will now not be pulled and therefore remained undefined. This is why there is a difference between undefined and null item states to identify the ones that we are expecting a valid state.

Based on my testing, this allows commands to be acknowledged as long as the relevant item state is valid while other item configured under the same interface may be invalid. However, it will generate an error when asking a device status for supported interfaces.

So using your test scenario as an example, Alexa will properly acknowledge when requesting to arm/disarm the device but it will return an error when requesting for the state as the alarm properties will be missing. Hopefully that makes sense.

digitaldan:
Awesome.  Having meaningful error messages is going to be a HUGE help for users trying to debug.   I'll give it a go in just a bit and test.  Is it ready to merge otherwise ?

jsetton:
It is ready to merge. Just need to know if we should move the troubleshooting section higher in the usage page. I am not sure users will be aware of it unless maybe we add a table of content at the top. What do you think?

digitaldan:
Yes, I agree it should be at the top for sure.

jsetton:
I have updated the usage page. The troubleshooting section is now above the setup and I also did some minor reshuffling in the process. Let me know what you think. Otherwise, this is ready to be merged.

Issue #165, opened 2019-03-06
[V3] Equalizer Controller Support


jsetton:
The [Alexa.EqualizerController](https://developer.amazon.com/docs/device-apis/alexa-equalizercontroller.html) interface provides directives that are used to set or adjust the equalizer bands and apply a sound mode to a smart entertainment device.

jsetton:
This is one is little tricky as there are different band properties part of one property. So we will need to introduce a concept of sub properties.

Here is my proposed sketch. The range configuration will be overlapping, due to interface definition, but we would use one of them assuming there are in sync.

```js
Group gReceiver         "Receiver" {alexa="Endpoint.Speaker"}
Number ReceiverBass     "Bass"     {alexa="EqualizerController.bands:bass" [range="-10:10",default=0]}
Number ReceiverMidrange "Midrange" {alexa="EqualizerController.bands:midrange" [range="-10:10",default=0]}
Number ReceiverTreble   "Treble"   {alexa="EqualizerController.bands:treble" [range="-10:10",default=0]}
String ReceiverMode     "Mode"     {alexa="EqualizerController.mode" [supportedModes="MOVIE,MUSIC,SPORT"]}
```

Issue #161, opened 2019-03-04
[v3] Using item commandOptions and stateDescription


digitaldan:
The openHAB core (formally ESH) added commandOptions and stateDescription to the item description in January, see https://github.com/eclipse/smarthome/pull/5131.  

This exposes the preset options for item commands/states that have existed for a while, but were not easily used by UI's using the rest interface, but are now.  For example, I have a String item which represents the current state of my logitech harmony remote.  If you send it a string, it will change to that current activity (like "Watch TV"). 

 I could see this being useful for Alexa channelControllers, modeController and possibly others.

Here's a snipit from from the /rest/items call:
```json
{
        "link": "https://myopenhab.org/rest/items/HarmonyHubMovieRoomActivity",
        "state": "PowerOff",
        "stateDescription": {
            "pattern": "%s",
            "readOnly": false,
            "options": [
                {
                    "value": "PowerOff",
                    "label": "PowerOff"
                },
                {
                    "value": "TiVo",
                    "label": "TiVo"
                },
                {
                    "value": "APPLETV",
                    "label": "APPLETV"
                },
                {
                    "value": "Game",
                    "label": "Game"
                },
                {
                    "value": "Roku",
                    "label": "Roku"
                }
            ]
        },
        "commandDescription": {
            "commandOptions": [
                {
                    "command": "PowerOff",
                    "label": "PowerOff"
                },
                {
                    "command": "TiVo",
                    "label": "TiVo"
                },
                {
                    "command": "APPLETV",
                    "label": "APPLETV"
                },
                {
                    "command": "Game",
                    "label": "Game"
                },
                {
                    "command": "Roku",
                    "label": "Roku"
                }
            ]
        },
        "editable": false,
        "type": "String",
        "name": "HarmonyHubMovieRoomActivity",
        "tags": [],
        "groupNames": []
    },
```

jsetton:
I am already using stateDescription option for ModeController implementation 😄 I wasn't aware of commandDescriptions.

digitaldan:
Could be useful too for thermostat modes when nothing is passed by the user:
```json
 {
                "link": "https://myopenhab.org/rest/items/F2_Hallway_HVAC_Mode",
                "state": "0",
                "stateDescription": {
                    "pattern": "%s",
                    "readOnly": false,
                    "options": [
                        {
                            "value": "0",
                            "label": "Off"
                        },
                        {
                            "value": "1",
                            "label": "Heat"
                        },
                        {
[...]

PR #168, opened 2019-03-14
Equalizer Controller Support

jsetton:
Fixes: #161
Fixes: #165

**Changes**
* Added ability to configure sub-property components to accommodate new capability
* Added ability to report a property under a different name than the one defined in discovery
* Added logic to determine thermostat binding 2.x parameter based on item channel metadata
* Added logic to determine mode range toggle friendly names based on item synonyms metadata
* Fixed log array object representation issue due to winston limitation
* Fixed discovery test directive request incorrectly including header correlation token
* Added schema validation to discovery test cases
* Updated validation json schema
* Moved mocha options into standard config file

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -5,7 +5,7 @@ Alexa is an intelligent personal assistant developed by Amazon and designed to r
 
 <p></p>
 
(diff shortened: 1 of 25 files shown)


digitaldan:
Looks good!

PR #169, opened 2019-03-15
Remove tag examples for metadata labels, add metadata examples in beg…

digitaldan:
…ining of page

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -37,7 +37,7 @@ The skill connects your openHAB setup through the [myopenHAB.org](http://myopenH
       * [Friendly Names Not Allowed](#Friendly-Names-Not-Allowed)
       * [Unit of Measurement Catalog](#Unit-of-Measurement-Catalog)
     * [Supported Group Metadata](#Supported-Group-Metadata)
-    * [Supported Labels](#Supported-Labels)
+    * [Supported Metadata Labels](#Supported-Metadata-Labels)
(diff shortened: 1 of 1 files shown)


digitaldan:
@jsetton I got some feedback that introducing the labels as new tags was going to be confusing to the larger OH community, which i agree with.  I originally put the tag usage examples in , since our labels happened to work that way, but i think that was a mistake as it both is confusing b/c they are not standard OH tags, and also it adds yet another way of doing things (3 in this case), which is never a good thing.   I don't think we need to restrict it in the code, but i think we should encourage  to use it it the Alexa meta data namespace.

jsetton:
I am fine with the change although we may want to have original v2 tags still documented, maybe in the v2 section, especially since these will interact with other similar integrations. I am not up to date if the other integrations have implemented new features using Alexa metadata but I think it is still worth leaving the relevant tags in the documentation.

digitaldan:
Sounds good, i just expanded the v2 support section with a table of v2 tags  -> v3 labels and added the example items from the homekit docs.

jsetton:
Looks good. I would add in that section that the v2 tags still need to be used to work with the other integrations such as HomeKit or Google Assistant. I am afraid that users would just convert them without realizing that they are removing the relevant configuration settings.

jsetton:
It would be nice if other integrations would at least support Alexa metadata v2 tags. That shouldn't be too hard to add 😄

digitaldan:
done

Issue #170, opened 2019-03-15
[V3] Alexa Skills Kit CLI Deployment Support


jsetton:
We should probably migrate the skill deployment to use the [ASK CLI](https://developer.amazon.com/docs/smapi/quick-start-alexa-skills-kit-command-line-interface.html). This will allow an all-in-one approach that will deploy the smart home skill along with its lambda function at the same time. Moreover, it makes it easier to deploy to various regions in one go and once we decide to implement Custom Intent Support (#164), the intent models will be able to be stored in the repository.

digitaldan:
👍

jsetton:
@digitaldan I didn't want to respond on the main beta thread. Have you try to use the ASK CLI beta test subcommands? Maybe the [list beta testers](https://developer.amazon.com/docs/smapi/ask-cli-command-reference.html#list-beta-testers-subcommand) subcommand might still return a list of users.

digitaldan:
It looks like when the beta is over, it completely wipes out the previous beta test.  I did not try the skills api (have not hooked it up to beta) , but in the web UI, everything was gone.

jsetton:
You don't need to hook it up. The api subcommands I mentioned above aren't related to the skill deployment. You just need to specify the skill id when calling the command. It is worth a shot but if you don't see in the web UI, the underlying data is probably gone as well.

PR #175, opened 2019-03-24
Alexa Skills Kit CLI Deployment Support

jsetton:
Fixes: #170 

Some of the documentation/configurations will need to be updated once v3 branch is merged into master.


jsetton:
If you want to recycle your current development skill in order to prevent duplicate devices, you just need to add your current skill id in the ask config file before running the deploy command.

https://github.com/openhab/openhab-alexa/blob/073af286bbf99ffacc48e7a4a7d4b2131e398ba1/.ask/config#L4

jsetton:
@digitaldan What do you think about this PR?

digitaldan:
Wow, thats awesome, going to give it a try today on my instance , but i think this is great.

jsetton:
Keep in mind that I didn't include the [other region deployment](https://developer.amazon.com/docs/smapi/skill-manifest.html#regions) configuration in the skill settings as (1) I am not sure how you deployed the official skill and (2) I don't think the folks that will be deploying the skill for private usage need the multi-region feature.

digitaldan:
I will look at it.  Right now i am deploying the skill to prod using the node-lambda cli which does push to multiple regions from the .env file  like:
```AWS_REGION=us-east-1,eu-west-1,us-west-1```

**Comment by jsetton** [comment]
So to achieve that, you would update [skill.json](https://github.com/openhab/openhab-alexa/blob/82fc30f03b15941d097549c812532118dc5f4ba9/skill.json#L248) with:

```diff
diff --git a/skill.json b/skill.json
index 45c8fd8..32e7367 100644
--- a/skill.json
+++ b/skill.json
@@ -251,9 +251,25 @@
     ],
     "apis": {
       "smartHome": {
-        "endpoint": {
-          "uri": "alexa-openhab",
-          "sourceDir": "lambda/smarthome"
+        "regions": {
+          "NA": {
+            "endpoint": {
+              "uri": "alexa-openhab",
+              "sourceDir": "lambda/smarthome"
+            }
+          },
+          "EU": {
+            "endpoint": {
+              "uri": "alexa-openhab",
+              "sourceDir": "lambda/smarthome"
+            }
+          },
+          "FE": {
+            "endpoint": {
+              "uri": "alexa-openhab",
+              "sourceDir": "lambda/smarthome"
+            }
+          }
         },
         "protocolVersion": "3"
       }
```

The only difference is that, based on the [default Lambda regions](https://developer.amazon.com/docs/smapi/ask-cli-intro.html#lambda-region), the far-east function would be deployed to us-west-2 opposed to us-west-1 as you currently have it setup.

The way I have it setup in repository is to use default region defined by your environment or fallback to us-east-1 if not defined.

digitaldan:
>deployed to us-west-2 opposed to us-west-1 as you currently have it setup.
That was a typo on my part, i'm actually deploying to us-west-2

jsetton:
Not sure if you went through the reference documentation, but you can also manage the beta testing along with the validation/submit process. You can potentially automate adding beta testers 😄

digitaldan:
Hey @jsetton sorry about the long delay, i have been on holiday and now traveling for work non stop, I will get this merged as soon as i can.

jsetton:
No worries 😄

digitaldan:
I'm finally back home for a bit and was able to play with this today.  Was pretty seamless, although i ended up somehow blowing away my original alexa skill trying to re-use it, so i will need to run through that a few times before i do this on the openHAB amazon dev account ;-)  Otherwise lgtm.  I'm going to catch up with our last discussion about a dedicated binding this week.

jsetton:
> although i ended up somehow blowing away my original alexa skill trying to re-use it

What do you mean by that? The original skill should be overwritten by the cli tool but the main purpose is to prevent having a new skill with new devices.

digitaldan:
I was trying to reuse the skill id, and i'm not sure what exactly happened , but it lost the oauth info i had setup.  I'm sure i screwed something up, but i'll run through it a few times with another skill just to verify the process.

jsetton:
As long as you update the `skill_id` in the ask config file prior to first deployment, you should be fine. It gets tricky after that.

PR #181, opened 2019-05-24
Fix StepSpeaker to pass step value to openhab

fmeies:
The documentation for the Alexa.StepSpeaker Interface says:

[...] implement this interface for devices that make incremental discrete step adjustments to volume, where the range of volume isn't known. For example, StepSpeaker can adjust volume by a step of 5, but can't adjust the volume to 50 percent. If your device can set the volume based on any continuous value in a range, implement the Speaker interface instead. [...]

I think that the StepSpeaker interface is meant for simple remote control like devices that do not have a volume state but only support "3 steps up" or "2 steps down" directives. Therefore I changed the adjustVolume function in stepSpeaker.js to only pass the incremental value to the device instead of modifying the state of the device (since the device does not have a state). What do you think?

Signed-off-by: Frank Meies <19324766+fmeies@users.noreply.github.com>

### DIFF
--- a/lambda/smarthome/alexa/v3/stepSpeaker.js
+++ b/lambda/smarthome/alexa/v3/stepSpeaker.js
@@ -7,7 +7,6 @@
  * http://www.eclipse.org/legal/epl-v10.html
  */
 
-const log = require('@lib/log.js');
 const AlexaDirective = require('../directive.js');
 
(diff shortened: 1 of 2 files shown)


fmeies:
This PR is covered by #182. Closed.

PR #182, opened 2019-05-25
StepSpeaker Controller Volume Steps Command Value Fix

jsetton:
* Seperated volume property schema to differentiate the two speaker capabilities
* Added support for default volume step increment metadata parameter
* Improved context response for non-reportable properties
* Improved invalid directive error handling when relevant property map settings not defined
* Minor code and documentation cleanup

Signed-off-by: jsetton <jeremy.setton@gmail.com>

### DIFF
--- a/lambda/smarthome/alexa/config.js
+++ b/lambda/smarthome/alexa/config.js
@@ -166,7 +166,7 @@ module.exports = Object.freeze({
       'category': 'SPEAKER',
       'properties': [
         {'name': 'muted', 'schema': 'muteState', 'isReportable': false, 'isSupported': false},
-        {'name': 'volume', 'schema': 'volumeLevel', 'isReportable': false, 'isSupported': false}
+        {'name': 'volume', 'schema': 'volumeSteps', 'isReportable': false, 'isSupported': false}
       ]
     },
     'TemperatureSensor' : {
@@ -451,6 +451,9 @@ module.exports = Object.freeze({
       'state': {
         'type': 'integer'
       }
+    },
+    'volumeSteps': {
+      'itemTypes': ['Number']
     }
   },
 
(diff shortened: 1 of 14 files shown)


jsetton:
@fmeies Good catch. That actually would explain why no properties response is expected for that interface. I should have done more testing with that specific interface.

Anyway, this PR includes your changes from #181. A new property schema needed to be created to enforce the item type for the StepSpeaker volume property to only be a number.

Additionally, I found out an undocumented feature for that interface (`volumeStepsDefault` property boolean being available in the directive payload), along with an issue where, StepSpeaker muted or volume directives would be sent to the skill even though we don't have one of them specified on the OH side, which is due to the limitation on the Alexa side as it is not processing enough discovery information for that specific interface.

@digitaldan let me know what you think.

lambda/smarthome/alexa/config.js
```diff
@@ -166,7 +166,7 @@ module.exports = Object.freeze({
       'category': 'SPEAKER',
       'properties': [
         {'name': 'muted', 'schema': 'muteState', 'isReportable': false, 'isSupported': false},
-        {'name': 'volume', 'schema': 'volumeLevel', 'isReportable': false, 'isSupported': false}
+        {'name': 'volume', 'schema': 'volumeIncrement', 'isReportable': false, 'isSupported': false}
```
digitaldan: small nit pick, is this really a increment or should we refer to it as steps, so  'volumeSteps', those steps may increment or decrement it.

jsetton: I kept going back and worth 😄 I went with `increment` because I have used that term as metadata parameter, but I agree with you and will change it.

digitaldan:
LGTM,  I want to check it out and run it  but need to step out for a few hours.  Once i get this merged i will proceed with merging the v3 branch into master.  

I believe this will close/negate PR #181  ?

jsetton:
> I believe this will close/negate PR #181

Correct as long as @fmeies is fine with the changes I made.

fmeies:
@jsetton @digitaldan Sure, go ahead, #181 is obsolete now with this PR. By the way: Thank you all for the tremendous job you're doing on this project!

PR #183, opened 2019-05-25
Merge V3 into master

digitaldan:


digitaldan:
Merging !

jsetton:
In case you missed it, the skill.json file needs to be updated. The resources folder was renamed from current master

jsetton:
@digitaldan I just noticed that you did a squash merge basically removing me as authored on any of the contributions I made to the v3 branch. Why not do a standard merge?

digitaldan:
Hi @jsetton , apologies, this was simply muscle memory for how we (openHAB) merges PRs. I used the built in merge button in the PR  page on github , which defaults to squashing commits.  I can revert the merge and choose not to squash, give me a few mins to do so.  

Regarding the skill file, i am going to update that based on our production config, i will do so probably tomorrow AM.

digitaldan:
FYi, i may need to do this from the command line, our repo actually does not support NOT squashing when using the web ui to merge

image

digitaldan:
Done!

jsetton:
> FYi, i may need to do this from the command line, our repo actually does not support NOT squashing when using the web ui to merge

Didn't intend to add more work here. Thanks very much for making the change.

> Regarding the skill file, i am going to update that based on our production config, i will do so probably tomorrow AM.

I am not sure what you want to do with the publishing files since there are now all embedded in the skill file. I was trying to think how we can keep using crowdin translation and potentially merge the results into the skill file. Do you want to keep that process going forward?

digitaldan:
The current translated descriptions we have are out of sync with the crowdin files, the review process was a never ending back and forth with amazon reviewers in each country and i ended making modifications  outside of crowdin.  My plan is to clone our current skill with ASK, which will pull down the current descriptions exactly. 
> I was trying to think how we can keep using crowdin translation and potentially merge the results into the skill file.

I had the same thought, using some type of hook or build script to generate the skill file. I'm hoping to get this releases out first without updating the descriptions, as that triggers a very lengthy review process.  That gives us time to fix the crowdin process and write a better, more v3 oriented description that can be translated on our own time.

jsetton:
> I had the same thought, using some type of hook

You just hit the keyword here. So we can use the [pre deploy hook script](https://github.com/openhab/openhab-alexa/blob/master/hooks/pre_deploy_hook.sh) to accomplish that. I just checked the ask-cli code and this is the first thing it runs when calling deploy command as long as the skill file has a defined skill name (so no issue here) and it reads the skill file against after the hook script ran.

The only thing I am not too familiar with is how crowdin is doing transaction on a json file. It seems that it supports that format. Do you think it would be able to handle the structure below?

https://github.com/openhab/openhab-alexa/blob/e8a2a27680f4694159b64127eab7ca3e3564c571/skill.json#L96-L111

PR #186, opened 2019-05-26
Skill Manifest Dynamic Localization Resources Support

jsetton:
I added an ask-cli pre-deploy task running a node tool script to update the skill manifest file, every time the `deploy` command is called, with localized resources that can be dynamically generated by crowdin.

I was able to test successfully the deployment part. However, the crowdin integration would still need to be tested as I am not 100% sure if the current file format can be used.


jsetton:
@digitaldan let me know what you think about this one.

jsetton:
So I just noticed you just submitted a PR for the skill file (#185). I saw you added the different api regional endpoints to deploy for production. One issue I can see is that if we keep it that way in the repository, everyone would be unnecessarily deploying their skill in the different regions. How about adding a step in the update skill manifest script part of this PR that would look for an environment variable?

Basically if the variable is set to production, then update the skill with necessary endpoints, otherwise leave as is. What do you think?

digitaldan:
Looking now, thanks!   Earlier I was looking at crowdin json support, https://support.crowdin.com/file-formats/json/  , it seems they do not support arrays ?  Have you solved for that?

If that is the case (which i have no confirmed) One thought i had was to do something like
```json
"examplePhrases1": "Alexa, schalte Küche ein",
"examplePhrases2":"Alexa, stelle Schlafzimmer auf 50 Prozent",
"examplePhrases3": "Alexa, reduziere Temperatur um 5 grad"
```

which would be translated to:

```json
  "examplePhrases": [
    "Alexa, schalte Küche ein",
    "Alexa, stelle Schlafzimmer auf 50 Prozent",
    "Alexa, reduziere Temperatur um 5 grad"
  ]
```

jsetton:
> Looking now, thanks! Earlier I was looking at crowdin json support, https://support.crowdin.com/file-formats/json/ , it seems they do not support arrays ? Have you solved for that?

I created a test project and loaded a test json file. It seems that it is able to parse the array per the screenshot below. I am more concerned having the url properties available. I am not familiar with crowdin. Will these be skipped?

image

jsetton:
Actually, I can see you can hide them. So it seems that it would work unless I am missing something.

image

jsetton:
> Basically if the variable is set to production, then update the skill with necessary endpoints, otherwise leave as is.

I just committed a change adding this support. Let me know what you think.
```
ASK_ENV=production ask deploy
```

jsetton:
Re: Crowdin, I was able to successfully generate a build, manually uploading the files under a test project and adding the relevant translations including Japanese and Portuguese (Used Google Translator for these two).

[Test Build Archive](https://github.com/openhab/openhab-alexa/files/3221695/test.zip)

digitaldan:
Thats awesome!  Looking at this now.

digitaldan:
lgtm, i'm going to merge this in and wire up our existing crowdin account

digitaldan:
Well, nothing to wire up :-)  Crowdin just sinked with master and updated our account with the whats in the repo. Very cool.

jsetton:
Nice! Which translated languages do you have setup?

PR #189, opened 2019-05-27
Update resource files based off current values in the production skill

digitaldan:


jsetton:
Not sure if you saw my comment on the icon URL. What is the reason behind using the current URL of the icon hosted by Amazon instead of the one pointing to this repository?

digitaldan:
I used the ASK clone function to pull down the current info and copied those into these files.  I was thinking the same thing before i merged which is why i closed it,  didn't see you comment until just now.

PR #192, opened 2019-05-27
Skill Manifest Environmental Variable Placeholder Support

jsetton:
@digitaldan I noticed that you had updated the skill instructions in a previous PR that wasn't merged. Added this in case it was missed.

### DIFF
--- a/skill.json
+++ b/skill.json
@@ -10,7 +10,7 @@
       "distributionCountries": [],
       "isAvailableWorldwide": true,
       "distributionMode": "PUBLIC",
-      "testingInstructions": "Provision an openHAB 2.x Server, install the 'demo items', configure, enable the openHAB Cloud Connector binding, create a myopenhab.org account and set your servers UUID & secret, say 'Alexa, Discover my devices'. You can then see the demo device states being controlled by Amazon Echo via the openHAB Web UI.",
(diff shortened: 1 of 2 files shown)


digitaldan:
i checked that change  in then realized that the openhab demo user's password was contained in the review instructions, so i reverted/deleted it. i thought i had pushed an update with just the removed password, but maybe i forgot to push it, or submit a PR (i'm juggling multiple projects this weekend) .  In any case, i quickly changed the demo password, so the one in this file won't actually work.  Have not given much thought to how to solve this issue.

jsetton:
No worries. I just open this PR as a reminder. Do you want me to update it removing the password or should I just close it?

jsetton:
> Have not given much thought to how to solve this issue.

We could use an environmental variable with the password that will get added to the skill manifest. I will update this PR with my thought.

jsetton:
So I added support for placeholders based on environment variable. In the testing instructions case, you would use based on the example below.

```
ASK_TESTING_PASS=<password> ask deploy
```

Let me know what you think. It might be over engineered maybe 😄 The placeholder should be named the same as environment variable without `ASK_` prefix (e.g. `ASK_TESTING_PASS` => `%TESTING_PASS%`)

digitaldan:
I like it!

PR #196, opened 2019-05-28
Set skill distribution mode based on deploy environment

jsetton:
Only use public distribution mode for production deployment, otherwise set it as private by default.

```
ASK_ENV=production ask deploy
```

### DIFF
--- a/skill.json
+++ b/skill.json
@@ -9,7 +9,7 @@
       },
       "distributionCountries": [],
       "isAvailableWorldwide": true,
-      "distributionMode": "PUBLIC",
+      "distributionMode": "PRIVATE",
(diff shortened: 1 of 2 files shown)


jsetton:
Let me fix that one.

digitaldan:
Sounds good, I have also granted you maintainer rights , so you can approve pull requests, write to master , etc, etc.   I hate being the bottleneck here :-)

jsetton:
> Sounds good, I have also granted you maintainer rights , so you can approve pull requests, write to master , etc, etc. I hate being the bottleneck here :-)

Thank you very much!

> Let me fix that one.

It's fixed. I just wanted to know if you liked the idea. The thinking behind is that even though someone tries to go through the publishing process it would end up just being a private skill.

digitaldan:
lgtm!

PR #200, opened 2019-05-31
Add PlaybackStateReporter interface support

jsetton:
For some unknown reason, the playback support is separated in two interfaces: one for commands and the other one for the current state. The former was already integrated under PlaybackController. This PR adds support for the latter.

https://developer.amazon.com/docs/device-apis/alexa-playbackstatereporter.html


digitaldan:
LGTM.

PR #206, opened 2019-06-08
Log level updates

jsetton:
* Decrease log level on response statements
* Set default log level to info


jsetton:
@digitaldan I changed the default log level to info to decrease the amount of logs generated especially during on discovery requests. Let me know what you think.

digitaldan:
lgtm, thanks!

Issue #195, opened 2019-05-28
[v3] Skill Manifest New Features Description


jsetton:
Do we want to use the [updatesDescription](https://developer.amazon.com/docs/smapi/skill-manifest.html#locales) skill manifest locales property to advertise the new features added in v3?

jsetton:
@digitaldan not sure if you saw this one. I wanted to know if you want to includes this one. It would be a good way to advertise the new v3 skill.

PR #208, opened 2019-06-08
Add skill manifest updates description

jsetton:
image

image

Fixes: #195


jsetton:
@digitaldan Let me know what you think about this one. We may have to wait until we agree on the final docs url. (#109)

digitaldan:
lgtm, since this is the current doc url, i think we should go with that.

PR #225, opened 2019-07-02
Add rest ssl client certificate auth support

jsetton:


jsetton:
@oeiber I believe you attempted to submit a change related to client certificate auth. I liked the idea so I implemented it. The only difference is that I enforced the use of PFX/P12 formatted certificate instead of traditional certificate files as this is what is most commonly used for client authentication and is in line with the Android openHAB app implementation.

Hopefully that would cover your use-case. Feel free to test and let me know if you are running into any issues.

oeiber:
@jsetton sounds good ;-) thank you!

oeiber:
works perfectly!
will it also be merged to the master branch?

jsetton:
@oeiber thanks for testing!

@digitaldan Can you go over this change in case I missed something? I added a change where the default OH base URL is now myopenhab.org. So technically, the `config.js` file wouldn't be necessary going forward for production releases.

digitaldan:
LGTM 👍

PR #227, opened 2019-07-08
Skill requires regions to contain an "endpoint" object

digitaldan:

### DIFF
--- a/tools/updateSkillManifest.js
+++ b/tools/updateSkillManifest.js
@@ -85,9 +85,8 @@ function setApiEndpoints(schema) {
   }
   if (process.env.ASK_ENV === 'production') {
     schema.manifest.apis.smartHome.regions = {};
-    REGIONS.forEach(region => {
-      schema.manifest.apis.smartHome.regions[region] = {endpoint: schema.manifest.apis.smartHome.endpoint}
-    });
+    REGIONS.forEach(region =>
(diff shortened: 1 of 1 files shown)


digitaldan:
@jsetton small tweak to make ask deployment happy when dealing with regions, otherwise everything else worked!

jsetton:
Good catch. Why not just initialize and set the object value at the same time?

```diff
diff --git a/tools/updateSkillManifest.js b/tools/updateSkillManifest.js
index fd46f72..28b6458 100644
--- a/tools/updateSkillManifest.js
+++ b/tools/updateSkillManifest.js
@@ -83,7 +83,7 @@ function setApiRegionalEndpoints(schema) {
   if (process.env.ASK_ENV === 'production') {
     schema.manifest.apis.smartHome.regions = {};
     REGIONS.forEach(region =>
-      schema.manifest.apis.smartHome.regions[region] = schema.manifest.apis.smartHome.endpoint);
+      schema.manifest.apis.smartHome.regions[region] = {endpoint: schema.manifest.apis.smartHome.endpoint});
   } else {
     delete schema.manifest.apis.smartHome.regions;
   }
```

digitaldan:
>Good catch. Why not just initialize and set the object value at the same time?

Yes, thats a better solution :-)

PR #234, opened 2019-08-10
Code refactoring capabilities & property map

jsetton:


jsetton:
@digitaldan let me know what you think on this one. I have converted the property parameter normalization switch case to a function object and I have ordered alphabetically similar code for determining capability interfaces. This is basically for clarity and to keep the size of a specific function to a manageable size.

digitaldan:
LGTM 👍

Issue #242, opened 2019-08-15
Large Item Collection Discovery Lambda Timeout Fix


jsetton:
Doing one big api call pulling every group members recursively is causing the discovery process to fail on large item collection as the lambda times out waiting for the response back.

The easy route would be to increase the lambda function timeout but that could be costly in the end. However, this could be just temporary until the logic to generate the list of discovered device is moved to the binding.

The other options would be to do an initial flat query similar to how it was done in v2 and only query endpoint-enabled groups when necessary. One downside with this approach will depend on how many of these groups a specific user will have, although we could have these calls run concurrently. Additionally, we should probably limit the output to a list of fields we use in the skill.

@digitaldan what do you think?

digitaldan:
Can we up the time of the request timeout?  I thought home skill kits had a limit regardless of the lambda timeout.  if we could up it just for the discovery request that great, but since it's on every request , yes it could be a cost issue. 

On the other hand, i don't see a ton of these, and they appear to only be on disco requests, so it may not be a problem, I'm actually a bit on the fence.  I'm inclined to up the timeout to 15 or 20 seconds  (if we can) and see if this solves our issue for a while.  

Shall i do this or would you like to?

jsetton:
> I thought home skill kits had a limit regardless of the lambda timeout.

I forgot about that part. We have 10 seconds to response to an Alexa request. This is probably why ask-cli sets the default lambda timeout to that value. Anything longer than that would require an immediate deferred response followed by an asynchronous one sent to the event gateway once available.

So it looks like we need to go for the latter option unless you have another idea?

digitaldan:
Let me think on it a bit this morning, i want to flush out an idea or 2 as a short term workaround until we have a longer term option.

digitaldan:
One idea for workaround today.  Just spitballing here, so not sold on anything. 

Supported deferred response just for Discovery requests.  Our skill would invoke another lambda on discovery that would perform the OH rest call with a timeout of 30 seconds and send the response it to the event gateway when it completes.   This avoids having to create a binding on OH which will take some time, and limits our exposure to longer running lambdas to just disco requests.  We would need to bring back your branch where we store in dynamo ( i think it was dynamo?) the user tokens .

jsetton:
> Supported deferred response just for Discovery requests

[...]

PR #243, opened 2019-08-16
Large Item Collection Discovery Lambda Timeout Fix

jsetton:
* optimized rest items query by disabling recursive parameter
* limited rest items query output to a list of fields use by skill

Fixes: #242

### DIFF
--- a/lambda/smarthome/alexa/propertyMap.js
+++ b/lambda/smarthome/alexa/propertyMap.js
@@ -369,9 +369,9 @@ class AlexaPropertyMap {
         // Append component name if property components required
         const propertyName = matches[2] + (settings.property.components ? ':' + matches[3] : '');
         const component = matches[3];
-        const properties = propertyMap[interfaceName] || {};
+        const properties = Object.assign({}, propertyMap[interfaceName]);
         const property = {
-          parameters: item.metadata.alexa.config || {},
+          parameters: Object.assign({}, item.metadata.alexa.config),
           item: {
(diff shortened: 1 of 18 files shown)


digitaldan:
I'll give it a try, but LGTM .  Pleasantly surprised how small the actual change was to the logic.  Nice job!

jsetton:
I just made an improvement by moving the group member filtering part only for endpoint-enabled groups along with the conversion. Make sure to get the latest update when testing.

digitaldan:
Something is not working with my environment which is causing discovered items to not show up.  I don't think its related at all to the changes in code but something with my setup i need to debug.  

I did however manage to run both the production skill and this version and diff'd  the response message.  The only difference in the 5334 lines of JSON was the messageId , which is of-course expected.  So I feel good this change.  I  need to turn in , but will pick this back up in the AM.  If you are comfortable with it i am good to push to production.

jsetton:
I am happy with the change. I tested as well on my end and none of my already discovered devices went offline which is a good sign 👍 It would be good if we could get the logger config changes included as well if you are fine with them.

Let me know if you need help pushing to production. For reference, in case you didn't know, you can simply update all production lambda functions only, using the ask-cli target parameter.

```
ask deploy -t lambda
```

digitaldan:
Sounds good, I take a look at the logger changes, but i'm sure they are fine.

> using the ask-cli target parameter.

yep, thats how i have been pushing the updates!

Issue #240, opened 2019-08-15
Winston Log Missing Lambda Request Id


jsetton:
The Winston logs are missing the Lambda request id which makes it hard to troubleshoot when trying to get all the logs related to a specific id.

According to this [issue](https://github.com/winstonjs/winston/issues/981), it is not writing to `console.log` and it is apparently not configurable in the standard console transport.

It seems the only solution currently is to create a [custom transport](https://github.com/winstonjs/winston#adding-custom-transports) that would write directly to `console.log`.

jsetton:
Ended up finding a way to configure the standard console transport using [this](https://github.com/winstonjs/winston/issues/1305#issuecomment-387634897) recommended solution.

PR #244, opened 2019-08-16
Update logger config format allowing lambda request id to be included

jsetton:
Previous Lambda log format:
`info: ...`

New Lambda log format:
`2019-08-16T05:47:37.452Z	33792eac-2485-45eb-aad4-8846fd8e4fc9	INFO: ...`

Fixes: #240

### DIFF
--- a/lambda/smarthome/lib/log.js
+++ b/lambda/smarthome/lib/log.js
@@ -13,18 +13,38 @@
 
 const { createLogger, format, transports } = require('winston');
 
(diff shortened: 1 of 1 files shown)


jsetton:
@digitaldan let me know if you are ok with this change. It will increase the size of each line but since the logs are only kept for a month, this should be fine. Anyway, the format is in line with how it was in v2 skill.

digitaldan:
👍

Issue #253, opened 2019-08-19
Discovery items type error failure


jsetton:
Per the skill logs, several users discovery requests are failing due to the error below

```
error: discover failed with error:items.forEach is not a function
{
    "stack": "TypeError: items.forEach is not a function\n    at Promise.all.then (/var/task/alexa/v3/discovery.js:45:13)\n    at <anonymous>\n    at process._tickDomainCallback (internal/process/next_tick.js:228:7)"
}
```

After enabling debug mode, the issue seems to be related to the items data value being a string because the request call seems to only have read/received a partial response preventing the expected JSON object from being parsed properly.

The example based on the reference below isn't that considerably big in size, although it's just partial data, and the Lambda execution time for that request was about 5 seconds.

[Lambda request log reference](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=/aws/lambda/openhab-alexa-v3;stream=2019/08/16/%5B$LATEST%5D64be4d08d373457fa4a1b2468f58d794;refid=34922537345580123117038527377609257753437664721776476160;reftime=1565980734486)

jsetton:
@digitaldan I moved this one into its own issue. I am seeing this error being generated a couple times every hour.

Anyway, I did some research and I found a [possible solution](https://stackoverflow.com/a/47822535) including `Connection: keep-alive` to the request header.

I did some testing and it looks like it is related to having the OH server behind a reverse proxy such as nginx. In that setup, the connection setting seems to be defaulting to `close`, if not specified, which would explain why the network connection could get prematurely closed.

Do you see any reason for not making this change?

```
REQUEST onRequestResponse https://myopenhab.org/rest/items?fields=editable%2CgroupNames%2CgroupType%2Cname%2Clabel%2Cmetadata%2Cstate%2CstateDescription%2Ctags%2Ctype&metadata=alexa%2Cchannel%2Csynonyms 200 { server:
'nginx',
  date: 'Mon, 19 Aug 2019 16:41:07 GMT',
  'content-type': 'application/json',
  'transfer-encoding': 'chunked',
  connection: 'close' }
```

digitaldan:
If that works i don't see a problem with it. I'm not sure why it works :-) afaik 'keep-alive' is used so the underlying TCP connection can be used multiple times by different requests and save having to create and tear down a TCP connection over and over for each request, browsers would be painfully slow without it.  

[...]

PR #261, opened 2019-08-25
Add v2 backward compatibility

jsetton:
* restructured project to clearly separate v2/v3


jsetton:
@digitaldan I decided to add v2 support back in the skill after reading [this documentation](https://developer.amazon.com/docs/smarthome/smart-home-skill-migration-guide.html#ensure-smooth-customer-migration) which stipulates that the skill should handle both versions until the skill no longer receive any v2 requests.

Also looking at the live skill logs, there are a ton of payload version errors still and these are for actual user requests to control their smart home devices. So rather force these users to spend the time to figure out their issue, we might as well make the experience as smooth as possible 😄

Anyway, let me know what you think. I made a minor project restructuring and added a few minor changes including having rest requests accept gzip compression if available which I think should be enabled with the myopenhab cloud connector #253.

On a side note, the state field is still being pulled in this one. I am hoping that having the request data compressed would resolve that matter.

digitaldan:
So given the size of the PR i will have to take your word on this one :-)  I can't easily tell what code you had to change and what was pulled in from the v2 connecter.  I hope it was not as much work as it looks.

digitaldan:
Nm my previous comment, i was not looking at the PR correctly and seeing files renamed vs added.  Changes looks good to me, super impressed at the v2 test classes you wrote, that does seem like a lot of work!   Thanks!

jsetton:
> I hope it was not as much work as it looks.

Not too much 😄 I had to cleanup and update a bit of the v2 ohConnector to integrate in the current framework. Most of the change is from the folder restructure. I think it make sense and was needed anyway.

> super impressed at the v2 test classes you wrote, that does seem like a lot of work

That part was actually the easier one. We already had the [test cases](https://github.com/openhab/openhab-alexa/commit/09945aa158fa7325217f24165aa59b21d6ceb96c) 😄

jsetton:
So I just pushed #265 to the live skill and will wait a day before merging and pushing this one to prod. That way we can see if the discovery performance changes improve the number of users that weren't able to have their devices converted to v3 because of the timeout issue.

digitaldan:
👍

digitaldan:
How did #265 go?  Do we need to push this one now?

jsetton:
So I just opened #269 related to #265. I think we should resolve that one before merging this PR. Unless you think otherwise.

Issue #257, opened 2019-08-20
Discovery request timing out due to image item state


jsetton:
For [users](https://community.openhab.org/t/alexa-smart-home-skill-v3-is-now-live/79848/131) that have image items, their discovery request is timing out due to overall size of the rest items query response which is pulling the raw data stored in the image item state. As a workaround, commenting these items resolves the issue.

Ideally, we would filter item types in the items query but the rest api only allows to filter based on one item type per request.

Another option would be to filter out the state field. We are currently using the state field only to determine the unit for item types supporting unit of measurement. This information can technically be retrieved in most cases from the item state presentation unless not part of the item label definition.

Should we enforce the item state presentation definition and forgo the logic using the state? One important point is that the state presentation is free form opposed to the state having a defined format. This mean the state is technically more reliable.

**Unable to determine unit**

- **item**
```
Number:Temperature TestTemperature "Test Temperature"
```

- **rest**
```
{
  "link":"https://myopenhab.org/rest/items/TestTemperature",
  "state":"60.0 °F",
  "editable":false,
  "type":"Number:Temperature",
  "name":"TestTemperature",
  "label":"Test Temperature",
  "tags":[],
  "groupNames":[]
}
```

**Able to determine unit**

- **item**
```
Number:Temperature TestTemperature "Test Temperature [%d °C]"
```

- **rest**
```
{
  "link":"https://myopenhab.org/rest/items/TestTemperature",
  "state":"20.0 °C",
  "stateDescription":{
    "pattern":"%d °C",
    "readOnly":false,
    "options":[]
  }
  "editable":false,
  "type":"Number:Temperature",
  "name":"TestTemperature",
  "label":"Test Temperature",
  "tags":[],
  "groupNames":[]
}
```

jsetton:
@digitaldan what do you think about this one? Is worth it to filter the state out from the rest get items query? Maybe there is another option I am not seeing?

digitaldan:
Uhg, I have never liked the concept of storing image data in an item, this being a prime example of why.  I'm torn b/c I don't think that its a widely used feature of OH, so my feeling is that not many people will have this problem.  Otoh, its obviously an issue we will continue to run into to.

I would personally punt on this one to see how widespread this is, possibly putting a note in the docs about this.  If you think not pulling the state  won't dramatically impact our ability to support UOM , then I'm good what whatever you decide.

jsetton:
[...]

PR #265, opened 2019-08-27
Improve discovery request performance

jsetton:
* removed state field from getting pulled in discovery request (#257)
* added UoM default value logic based on dimension and regional settings
* streamlined rest request options now accepting gzip compressed data (#253)
* separated config file user and pass properties to align with env variables


digitaldan:
lgtm 👍

PR #294, opened 2019-10-16
Revert mode range toggle controller latest changes

jsetton:
* Added back non-controllable property parameters requirement
* Rolled back regional availability to US only for now


jsetton:
My test setup got corrupted and I assume that some of the latest changes I added for mode, range and toggle controllers were working properly while they didn't.

Regarding the regional availability, Amazon has started to update some of their [documentation](https://developer.amazon.com/docs/device-apis/list-of-interfaces.html) but from my initial testing, it is not working for other languages as of yet. So, I decided to rollback the documentation updates to prevent any confusion and we should update when it is fully expanded on the Alexa side. I actually got too eager! 😄

Issue #283, opened 2019-10-04
Open Close Raise Lower Commands Semantic Extensions Support


jsetton:
Amazon just [announced](https://developer.amazon.com/blogs/alexa/post/49eff766-5d95-4faf-9205-2ab52ae8a437/announcing-updated-smart-home-building-block-apis-non-controllable-device-states-semantic-extensions-and-expanded-international-support) support for open, close, raise and lower commands via sementic extensions using mode, range and toggle controllers.

It seems Amazon is focusing on improving these controllers instead of creating specific ones that would have limited purposes. This means that we should probably start creating metadata labels for generic device using these controllers. For example, part of this enhancement, a blind-specific label should be included.

CrispinP:
This functionality would be really cool. 

"Alexa, open the blinds" :)

jsetton:
> This means that we should probably start creating metadata labels for generic device using these controllers.

Related to this point, the `CurrentHumidity` metadata label added in #292 is the first example.

jsetton:
I have pushed a PR to add this support. Below are examples on how to configure [semantic extensions](https://developer.amazon.com/docs/device-apis/alexa-discovery.html#semantics-object) on the mode, range and toggle controllers. Additionally, the `Blind` and `Door` metadata label were added along with the [HomeKit](https://www.openhab.org/addons/integrations/homekit/#item-configuration) `WindowCovering` tag as an alias of the former.

@digitaldan let me know what you think.

**Mode**
```
String Shutter "Shutter" {alexa="ModeController.mode" [category="EXTERIOR_BLIND", friendlyNames="@Setting.Opening", supportedModes="Up=@Value.Open,Down=@Value.Close", actionMappings="Close=Down,Open=Up,Lower=Down,Raise=Up", stateMappings="Closed=Down,Open=Up"]}
```

**Range**
```
Rollershutter Blind "Blind" {alexa="RangeController,rangeValue" [category="INTERIOR_BLIND", friendlyNames="@Setting.Opening", supportedRange="0:100:10", unitOfMeasure= "Percent", actionMappings="Close=0,Open=100,Lower=(-10),Raise=(+10)", stateMappings="Closed=0,Open=1:100"]}
```

**Toggle**
```
Switch Door "Door" {alexa="ToggleController.toggleState" [category="DOOR", friendlyNames="@Setting.Opening", actionMappings="Close=OFF,Open=ON", stateMappings="Closed=OFF,Open=ON"]}
```

PR #325, opened 2019-12-07
Add semantic extensions support

jsetton:
Fixes: #283

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -18,6 +18,7 @@ The skill connects your openHAB setup through the [myopenHAB.org](http://myopenH
   * [Concept](#concept)
     * [Single Items](#single-items)
     * [Group Items](#group-items)
+    * [Semantic Extensions](#semantic-extensions)
     * [Item Sensor](#item-sensor)
     * [Item State](#item-state)
     * [Item Unit of Measurement](#item-unit-of-measurement)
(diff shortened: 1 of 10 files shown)


digitaldan:
Very cool!  I have some zwave/somfy blinds i will hook up to this tonight and let you know.  Code looks great, but i will take a deeper look when i play with the functionality.  Definitely a requested feature.

jsetton:
@digitaldan Looking for some feedback on the semantic metadata parameters syntax I came up with. It looks pretty straight forward to me but wanted to hear your point of view. I kept it close to how it is defined in Alexa Skill API.

digitaldan:
Been testing this all day yesterday, Looks great! I say merge it!

jsetton:
Awesome! 👍 What about my other point related to the `Blind` metadata label? What are your thoughts?

digitaldan:
>Also, the Blind metadata label currently have the Lower and Raise action semantics as +/-10% adjustment range values. I am wondering if we should just have them match the Close and Open actions. 

i think that makes sense, some blinds only have on/off, fully up or down settings anyways.  The example usage shows how to overwrite that anyways with ` actionMappings="Close=100,Open=0,Lower=(+10),Raise=(-10)"` , correct?  So the default behavior works for everyone, and advanced configs can further customize behavior. 

>Another point, I noticed that the Google Assistant integration uses tag Blinds.

I say use the correct value, so `Blind`.

jsetton:
> i think that makes sense, some blinds only have on/off, fully up or down settings anyways. The example usage shows how to overwrite that anyways with actionMappings="Close=100,Open=0,Lower=(+10),Raise=(-10)" , correct? So the default behavior works for everyone, and advanced configs can further customize behavior.

This is correct. I will keep it as is then.

PR #356, opened 2020-03-29
Fix: Bulb-config in doc fixed

szilch:
It should be ColorTemperatureController

Signed-off-by: Stefan Zilch <st.zilch@gmail.com> (github: szilch)


szilch:
Hey dev community. I made a small fix in the documentation. There is an example on how to configure bulb for Alexa. I took me 2h in my openhab config to figure out that it should be: **ColorTemperatureController.colorTemperatureInKelvin** instead of ColorTemperature.colorTemperatureInKelvin :)

Cheers Srefan

PR #359, opened 2020-05-30
Update instructions

llamahunter:
Was missing the critical step of installing the required module dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,14 @@ To configure the server path and credentials, you will need to setup the applica
     ```
     $ git clone https://github.com/openhab/openhab-alexa.git
     ```
+2. Install nodejs dependencies
+    ```
+    cd lambda/smarthome
+    npm install
+    cd ../..
+    ```
 
-2. Deploy the skill and the lambda function in one step:
+3. Deploy the skill and the lambda function in one step:
     ```
     $ ask deploy
     Profile for the deployment: [default]
@@ -82,7 +88,7 @@ To configure the server path and credentials, you will need to setup the applica
     [Warn]: Skill api domain "smartHome" can not be enabled. Skipping the enablement.
     ```
 
-3. Setup skill account linking using the skill id displayed in previous step and your OAuth2 provider configuration:
+4. Setup skill account linking using the skill id displayed in previous step and your OAuth2 provider configuration:
     ```
     $ ask api create-account-linking -s <skillId>
     ? Authorization URL:  https://www.amazon.com/ap/oa
@@ -98,7 +104,7 @@ To configure the server path and credentials, you will need to setup the applica
     Account linking created successfully.
     ```
 
-4. Enable skill with account linking:
+5. Enable skill with account linking:
     * Go to your [Alexa skill console](https://alexa.amazon.com/spa/index.html#skills/your-skills/?ref-suffix=ysa_gw)
     * Click on the "openHAB" skill under the "Dev Skills" tab
     * Click "Enable" and go through the account linking process


jsetton:
@llamahunter Thanks for your contribution.

The ASK-CLI v1 pre-deploy hook script should have automatically installed the dependencies before being deployed. I am not sure why you had to do it manually.

https://github.com/openhab/openhab-alexa/blob/adeae1f2a188264b31824f05aa7755cdecf5ac70/hooks/pre_deploy_hook.sh#L23-L26

llamahunter:
It didn't. I got runtime errors on my lambda function about being unable to load modules. After going down into the lambda dir and running `npm install` there, then back up and `ask deploy` again, it worked.  I was using ask 1.7.23

jsetton:
> I got runtime errors on my lambda function about being unable to load modules.

Can you delete the `node_modules` directory from `lambda/smarthome` and `ask deploy --debug`? Also, can you provide the runtime error?

llamahunter:
Ah, I see there's a bunch of interesting grep and sed stuff.  Have you checked that that works as expected on BSD?  For instance
```
openhab-alexa % grep "sourceDir" ./skill.json | sed -E 's/.*:\s*"(\[^"\]+)".*/\1/g'
          "sourceDir": "lambda/smarthome"
```

I suspect you want that to just return `lambda/smarthome`?

llamahunter:
I think it's the `\s` for spaces in the sed line.  That's a gnu extension and doesn't seem to work in bsd.  I think you need
```
grep "sourceDir" ./skill.json | sed -E 's/.*:[[:space:]]*"([^"]+)".*/\1/g'
```

llamahunter:
It's also interesting because the current code *created* a directory at the top level called `"sourceDir":` (yes, with the quotes and colon).  I think that `npm install` must have created that when passed it as the first agrument to `install_dependencies`.

jsetton:
> I suspect you want that to just return `lambda/smarthome`?

This is part of the template hook scripts provided by the cli tool. It's meant to be generic for any project.

> Have you checked that that works as expected on BSD?

I didn't. It's been a while since I have used that OS 😄 You probably pinpoint the actual problem. Anyway, the plan is to update this project to support the ASK-CLI v2 which will deprecate these hook scripts.

> I think it's the \s for spaces in the sed line. That's a gnu extension and doesn't seem to work in bsd.

Feel free to update this PR to fix the `sed` command.

llamahunter:
That's provided by a cli template?  Surprising that no one else has encountered the bug, since a lot of people use MacOS.

Anyway, I will close this PR and open a new one to fix the sed expression to be POSIX.

jsetton:
> It's also interesting because the current code created a directory at the top level called "sourceDir": (yes, with the quotes and colon).

That's because if the `sed` command fails, `SOURCE_DIR` gets the full config line `"sourceDir": "lambda/smarthome"` instead of the actual directory. So `npm install sourceDir` gets called instead.

> That's provided by a cli template? Surprising that no one else has encountered the bug, since a lot of people use MacOS.

Actually, I just checked and it was actually changed to introduce source directory uniqueness as the template run `npm install` multiple times unnecessarily when deployed to multiple regions like we do with the live skill. The is the command from the default template.

```sh
grep "sourceDir" ./skill.json | cut -d: -f2 |  sed 's/"//g' | sed 's/,//g' | while read -r SOURCE_DIR; do
```

> Anyway, I will close this PR and open a new one to fix the sed expression to be POSIX.

In all, good catch! 👍

PR #364, opened 2020-07-10
Add reference to config example with channel metadata

LeeC77:
This took me a while to find on the Forum and its not documented in the openHAB Amazon Alexa Smart Home Skill.
https://community.openhab.org/t/tagging-devices-for-alexa-support/98155/3


jsetton:
Thanks for your contribution. Since the item metadata concept isn't specific to the Alexa skill, I think what you highlighted should be documented at the OH level probably in [here](https://www.openhab.org/docs/concepts/items.html#item-metadata). That way other integrations using metadata (e.g. HomeKit or Google Assistant) could refer to that information as well.

PR #376, opened 2020-11-17
Update USAGE.md

BigGeorgeTx:
I was troubleshooting some devices that Alexa wouldn't discover.  Ultimately the problem was that I missed the uppercase C in PowerController.  It took me a while to figure out that I had entered Powercontroller.powerstate.  Once I figured it out and changed it, the devices were quickly discovered.  I'm proposing adding this to the ### Device Not Found section of #Troublshooting


jsetton:
This was included in the latest change.

https://github.com/openhab/openhab-alexa/blob/main/USAGE.md#device-not-found

Issue #391, opened 2021-04-18
Separated by spaces channels aren't discovered by alexa


Trysupe:
Hey there=)

Today I found an interesting issue with the alexa binding: 

When adding the alexa channels to the item there are no spaces allowed between the channels themselves. Let me show you:

This works:
```
Color Sprachassistent_Licht_Bett "Bett" {alexa="PowerController.powerState,BrightnessController.brightness,ColorController.color" [category="LIGHT"]}
```
This doesn't:

```
Color Sprachassistent_Licht_Bett "Bett" {alexa="PowerController.powerState, BrightnessController.brightness,ColorController.color" [category="LIGHT"]}
```

When using the second configuration alexa can't control the brightness of the item.

These spaces are automatically added (in my case) by the openhab alignment extension of vscode. I opened an [issue there as well](https://github.com/MaxBec/openHAB-Alignment-Tool/issues/59). However I thought it might be desirable to have the alexa binding also accept channels separated by spaces.

jsetton:
> However I thought it might be desirable to have the alexa binding also accept channels separated by spaces.

Agreed 👍

Issue #390, opened 2021-04-16
Separate Label


delid4ve:
For naturally spoken commands Alexa requires names such as Lamp or Light.  As it currently stands the skill utilises the openHAB label for this.

There are issues either way with this..

Example:
* All Lamps named as Lamp
    * openHAB ui everything indistinguishable and hard to manage
    * Alexa - naturally spoken, works correctly in groups

* Lamps named specifically such as Lounge Lamp
    * openHAB ui clean
    * Alexa - naturally spoken doesn’t work correctly as standard
    * Time consuming renaming in Alexa app to just Lamp
        * If corruption on Alexa profile this has to be done every time

My proposal is that a separate customLabel tag is added to the metadata for channels.  In order to provide backwards compatibility if the customLabel tag is not present then it defaults to the openHAB label.

Other workaround:
I don’t know how this applies to v3 ui but at present on 2.5.13 I do the following:
Create a separate group (if not already using advanced endpoints) for all channels required to be exposed and use the advanced meta on the item itself.

A small snippet of this can be found here:
https://community.openhab.org/t/evolvement-of-items-in-ohx/120315/11

jsetton:
> My proposal is that a separate customLabel tag is added to the metadata for channels. In order to provide backwards compatibility if the customLabel tag is not present then it defaults to the openHAB label.

We could add an optional `name` metadata parameter that would take precedence over the item label if specified, similar to how it is implemented with the Homekit integration.

To be clear, based on the forum discussion you referred, there is no support for different names for a given device on the Alexa end as of yet, compared to Google Assistant which seems to allow it.

PR #394, opened 2021-07-09
Major code refactoring

jsetton:
Fixes: #390, Fixes: #391

**Changes**
* introduced new item configuration syntax using device metadata concept
* added new device capabilities for camera, door/window covering, fan, networking and vacuum
* improved security arm state error handling
* added input custom mappings support
* added playback stop and thermostat hold/fan capabilities support
* added new custom asset ids and semantic extensions
* added switch support to mode generic capability
* improved dimmer/rollershutter support for range value generic capability
* updated test suite increasing code coverage
* added support for ask-cli v2 using cloudformation deployer
* improved ci integration with automated production release deployment
* added github action to update repository locales on crowdin changes
* deprecated item sensor concept for decouple state support on specific capabilities
* removed item tag support for oh3.x
* removed alexa smarthome api v2 support


jsetton:
As side notes, this change includes the support for the [KeypadController interface](https://developer.amazon.com/en-US/docs/alexa/device-apis/alexa-keypadcontroller.html) but I wasn't able to get it working during my testing. I have a support ticket opened with the Alexa developer support team but I haven't got any response for couple weeks now.

Also, since I don't own an echo device with a display 😄 , I wasn't able to properly test the support for the [CameraStreamController interface](https://developer.amazon.com/en-US/docs/alexa/device-apis/alexa-camerastreamcontroller.html).

lambda/openhab/index.js
```diff
@@ -0,0 +1,250 @@
...
+    return this._request(options);
+  }
+
+  /**
+   * Returns item state formatted based on its state description pattern and type
+   * @param  {Object} item
+   * @return {String}
+   */
+  static formatItemState(item) {
+    const format = item.stateDescription && item.stateDescription.pattern.match(/%(?:[.0]\d+)?[dfs]/);
```
digitaldan, line 193:
```suggestion
    const format = item.stateDescription &&  item.stateDescription.pattern && item.stateDescription.pattern.match(/%(?:[.0]\d+)?[dfs]/);
```
Just testing now,  I was getting an error that item.stateDescription.pattern was undefined here when calling `match` , this seemed to fix it.

jsetton: Thanks I will include the fix. There are other instances where I made the assumption that the `pattern` is always present if `stateDescription` is available.

I made the same assumption with `options` but from my testing, it seems that this property is always present.

digitaldan:
Since your last commit, i have not seen any issues, this is using a pretty extensive amount of devices and configurations.   The "Blind" endpoint also work nicely.  I have not yet spent time getting the camera endpoints up, our IPCamera binding uses a IP port per camera, which really complicates things.  Otherwise this PR looks really really good!

jsetton:
> I have not yet spent time getting the camera endpoints up, our IPCamera binding uses a IP port per camera, which really complicates things.

That's the reason I mentioned about the complexity in the usage documentation. Using NGINX, you would configure a HLS stream as follow:

```nginx
server {
  listen 443 ssl;
  server_name cam1.myserver.com;

  ssl_certificate cam1.myserver.com.crt;
  ssl_certificate_key cam1.myserver.com.key;
  ssl_protocols TLSv1 TLSv1.1 TLSv1.2;

  auth_basic Restricted;
  auth_basic_user_file htpasswd;

  location / {
    proxy_pass http://127.0.0.1:54321;
  }
}
```

> Otherwise this PR looks really really good!

Thanks 😄 Are you good with the new syntax? If so, I will make an announcement on the community forum asking for some feedback and starting a beta test.

jsetton:
> Fyi, i just found an issue where one of my items was not discovered

`ToggleController.toggleState` doesn't support `String` item type. I am surprised it was getting discovered before. It should have been `ModeController.mode` instead. On a side note, what binding are you using on this item? I can't seem to find it on the openHAB addons website.

digitaldan:
Interesting, it was working before, so i never noticed the syntax error.  I'll try fixing it, thanks for the quick diagnosis!

>On a side note, what binding are you using on this item? I can't seem to find it on the openHAB addons website.

I have not submitted it yet 😁  Its a binding to control a Grandstream GDS 3710, http://www.grandstream.com/products/facility-management/facility-access-systems/product/gds3710  , i have one controlling my driveway gate .

jsetton:
> Interesting, it was working before, so i never noticed the syntax error.

Could it be it was a `Switch` item type when initially discovered? I can't see how it would be working with `ToggleController.toggleState` since that capability can only send `ON` or `OFF` commands.

digitaldan:
> Could it be it was a Switch item type when initially discovered?

it very well could have been

jsetton:
@digitaldan Amazon finally came back to me about the issue related to the [KeypadController interface](https://developer.amazon.com/en-US/docs/alexa/device-apis/alexa-keypadcontroller.html) I am not able to test. They seem to indicate that it only works with an Echo device with a display which I don't believe is the problem here.

Anyway, have you had a chance to test this one? If not, you could use the following test scenario:

#### Item Definition
```
Group gStreamingDevice "Streaming Device" {alexa="StreamingDevice"}
Switch Power "Power" (gStreamingDevice) {alexa="PowerState"}
Player Player "Player" (gStreamingDevice) {alexa="Playback"}
String Navigation "Navigation" (gStreamingDevice) {alexa="Navigation" [supportedKeys="UP,DOWN,LEFT,RIGHT,SELECT,INFO"]}
```

#### Utterances
*Alexa, turn on streaming device*
=> Command `ON` sent to `Power`
*Alexa, play*
=> Command `PLAY` sent to `Player`
*Alexa, show info*
=> Command `info` sent to `Navigation` (This didn't work for me)
*Alexa, scroll down*
=> Command `down` sent to `Navigation` (This didn't work for me)

PR #427, opened 2021-12-05
Avoid HTTP 302 Redirect on Root Rest Endpoint

csowada:
In OpenHAB 3.1 the endpoint is ``http(s)://openhab:80/rest/`` instead of ``http(s)://openhab:80/rest``.

In my environment, this behavior has caused an error. Openhab redirected from HTTPS to HTTP. This caused a "400 Bad Request" on nginx.

This tweak avoids that redirect. I don't know if this also works with openhab 2.x.

### DIFF
--- a/lambda/openhab/index.js
+++ b/lambda/openhab/index.js
@@ -152,7 +152,7 @@ class OpenHAB {
   getRootResource() {
     const options = {
       method: 'GET',
-      uri: '/rest',
+      uri: '/rest/',
       json: true
     };
     return this._request(options);


jsetton:
Thanks for your contribution. I see this is an issue when pointing the skill directly to an openHAB server.

Could you apply the same change to [each instance](https://github.com/openhab/openhab-alexa/blob/main/lambda/test/openhab.test.js#L278) of the testing suite? Also, make sure to signoff you committed.

Based on my testing, this only affects OH 3.x instances.

#### openHAB 3.x
```
curl -I http://127.0.0.1:8080/rest
HTTP/1.1 302 Found
Location: http://127.0.0.1:8091/rest/
Content-Length: 0
Server: Jetty(9.4.43.v20210629)

curl -I http://127.0.0.1:8080/rest/
HTTP/1.1 200 OK
Content-Type: application/json
Date: Mon, 06 Dec 2021 00:21:42 GMT
Transfer-Encoding: chunked
Server: Jetty(9.4.43.v20210629)
```

#### openHAB 2.x
```
curl -I http://127.0.0.1:8080/rest
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 1195
Server: Jetty(9.4.20.v20190813)

curl -I http://127.0.0.1:8080/rest/
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 1195
Server: Jetty(9.4.20.v20190813)
```

jsetton:
@csowada do you think you could make the changes I mentioned above? We unfortunately cannot merge this PR as is.

jsetton:
So I went ahead and merge PR #429 including this change. I wanted to get in the next beta release.

Thanks again for raising this issue. 👍

PR #542, opened 2022-12-28
Convert to es module

jsetton:
AWS Lambda started to [support ES modules](https://aws.amazon.com/about-aws/whats-new/2022/01/aws-lambda-es-modules-top-level-await-node-js-14/) from nodejs14 since the beginning of the year.

While looking at replacing the deprecated [`request`](https://www.npmjs.com/package/request) http client package, there is an increasingly amount of maintained packages that are going ESM support only. I ended making the conversion to test some of the latest http client packages such as [`got`](https://www.npmjs.com/package/got) or [`node-fetch`](https://www.npmjs.com/package/node-fetch). In the end, I settled with [`axios`](https://www.npmjs.com/package/axios) for the time being since it requires the least amount of changes in our codebase (Will be part of a separate PR). I am curious to see how the native Fetch API evolves.

### DIFF
--- a/lambda/.eslintrc.json
+++ b/lambda/.eslintrc.json
@@ -1,7 +1,7 @@
 {
   "extends": [
     "eslint:recommended",
-    "plugin:node/recommended",
+    "plugin:n/recommended",
     "plugin:prettier/recommended"
   ],
   "parserOptions": {
@@ -9,6 +9,6 @@
   },
   "rules": {
     "no-console": "warn",
-    "node/no-missing-require": "off"
+    "n/no-missing-import": "off"
   }
 }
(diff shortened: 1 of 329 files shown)


digitaldan:
Wow!

jsetton:
@digitaldan What do you think?

digitaldan:
I think moving to ES modules sounds great.

jsetton:
Great. Any objection if I merge this one?

My local skill is working without any issues. The conversion is most syntax related with some minor refactoring to account for how import/export works with ES modules.

digitaldan:
Merging sounds good! As you mentioned the changes look mostly syntax related, and a lot of them,  so  i can't add much in terms of a review.

